PR #10, opened 2020-03-23
Build time dependency checking on pybind module files

mdavis36:
These changes should allow the make system to rebuild pybind modules whenever a *MOD.py file changes or any of its imported python files.

It works by generating an initial list of dependency files at configuration time that are then checked for changes during the build.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ syntax: glob
 *.C
 *.log
 *.silo
+*_stamp.cmake
 *.job
 *.job.out
 job.out*
--- a/cmake/CMakeBasics.cmake
+++ b/cmake/CMakeBasics.cmake
@@ -22,18 +22,21 @@ function(instantiate _inst_var _source_var)
   foreach(_inst ${${_inst_var}})
     if(ENABLE_INSTANTIATIONS)
       foreach(_dim ${_dims})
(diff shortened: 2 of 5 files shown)


jmikeowen:
So if we change the set of files imported by a *MOD.py file, does it automatically regenerate the set of dependencies?  I'm guessing it needs to rerun cmake?

mdavis36:
@jmikeowen Correct if a file is added or removed it will rebuild the module (since the top parent file was changed), but if you make any changes later to that added file you will have to rerun cmake as make won't know the added file exists as a dependency. I don't know if there is a way to do build time dependency generation

jmikeowen:
Hmm.  We should check it to be sure.  This would represent losing capability that works in autoconf, and there ought to be some way to fix this.

mdavis36:
Okay, I'll look into this more then

mdavis36:
Okay so the latest changes now seem to handle everything we want to do. It works by generating a cmake file that contains a list of the python deps during configure time. During build time the cmake file is regenerated and if any changes to the dependency list have been made it is replaced. This then prompts cmake to rerun a configure stage automatically, to make sure the new file is tracked as a dependency during build time. @jmikeowen

jmikeowen:
This looks good to me -- I don't see any reason not to go ahead and merge it.

PR #14, opened 2020-05-12
CMake build time TPLs

mdavis36:
### Summary
This PR allows the cmake system to build TPLs as build time targets along with other changes and bugfixes:
- Build time Third party library builds as dependent targets.
- Moved CMake root directory to Spheral/.
- Fixed r-paths of built libs.
- TPLs build into the cmake build dir by default.
- Fixed configured files to generate to build dir and install later.
- Generate python stamp files in build tree.

### CMake Flags
- SPHERAL_INSTALL_DIR : Unchanged. This is where any built TPL's and Spheral libs will be installed.

- BUILD_TPL=On/Off : On by default, when off NO TPLs will be built, will search in default location for libs, otherwise user can define the TPL install directory to search.

- \<tpl-name\>_BUILD=On/Off : On by default unless BUILD_TPL=Off. Define if user does not want to build a TPL.

- \<tpl-name\>_DIR : absolute path to TPL install location. If this is not set, cmake will set to default/SPHERAL_INSTALL_DIR. This location can be set as a search location when not building or a specific Install location for the given TPL.

- ENABLE_CXXONLY : Only build C++ shared libs.

- ENABLE_STATIC_CXXONLY : Only build C++ static libs.


mdavis36:
@ptsuji This is the PR for the new CMake system

jmikeowen:
I still have to check out the latest, but I have one question -- are the stamp files (I think from PYB11) still being written to the source tree?

mdavis36:
@jmikeowen Nope they are generated into the build tree now.

jmikeowen:
Awesome!  I'll try to check it out soon...

jmikeowen:
This is looking good!

PR #20, opened 2020-07-28
Bugfix/squash warnings for RK

mdavis36:
# Summary
This PR is one of many in order to begin squishing out the flood of warnings generated when building Spheral. For this particular instance I focused on squashing all warnings when building the C++ libs for RK. Hopefully a lot of the fixes in this initial PR will make the next modules go faster.

### Mismatched sign comparison on for loops
Primarily found in for loops with auto i and when the comparative type was typically unsigned int - Fixed by declaring `auto i = 0u` instead of `auto i = 0`.

### Constructor member initializer list ordering
Mismatched ordering between class definitions and constructor initializer list ordering. - Simple fix, rearrange the class definition, or the member initializer list order.

### Unused Variables/Arguments
I tried to keep the integrity of the code where a potential argument could be descriptive by commenting it out. If the variable name was useless and only used for overloading specialization I removed it outright. Other times I wrapped it in SPHERAL_UNUSED macro when it was evaluated in a DBC macro that wasn't compiled. I tried to be as consistent as possible here.

### Unused typdefs
Removed.

### Unused functions
Some functions were throwing unused function warnings. These were in RK/ContinuityVolumePolicy.cc and RK/computeRKSumVolume.cc. I had a harder time suppressing these warnings as the are overloaded on the Dimension type and were defined in the anon namespace. For one of them I made the functions constexpr, the other I moved inside the class as private members. @jmikeowen please let me know if you see any issue with those changes.


mdavis36:
@jmikeowen Is Misha on the project, I can't seem to add him as a reviewer?

jmikeowen:
I checked, and it looks like I never put the CONTRACT_VAR statement in Spheral's DBC stuff, which was an oversight.  I'd suggest we rename the SPHERAL_UNUSED macro to CONTRACT_VAR, since I think that's clearer.  These variables are not unused, they're just only used by contract statements.

mdavis36:
@jmikeowen Sounds good, perhaps SPHERAL_MAYBE_UNUSED would have been a more appropriate name, but I'll go ahead and rename to CONTRACT_VAR.

jmikeowen:
Whoa, I didn't realize naming arguments in the function signature would also trip these warnings if those args are not used.  That's a bummer.

The only issue I see is I'd rather not make some of those former functions in the anonymous namespace into private methods.  That means if we change those at some point we can trip unnecessary recompiling, and philosophically those methods don't require anything from the class so they don't need to be members.

I'd suggest 2 possibilities:
- If they really are unused we can just comment 'em out.  We must have left them there in case we wanted to come back and use them again at some point.
- Does inlining those methods avoid the warning?  Probably not but I'm curious.

mdavis36:
@jmikeowen As we discussed offline the Hvolume functions ContinuityVolumePolicy have been pulled out of the class. Making the static inline has seemed to have fixed the warnings. 

This is ready to go if you're good with everything else.

PR #23, opened 2020-08-18
Wsl2 port

jmikeowen:
This branch is for testing out the build on Windows 10 (Ubuntu).  The only problem I ran into was the version of CMake didn't support the "TYPE LIB" syntax in the install command, so I changed that to "DESTINATION" in the maneos third party lib build.


MishaZakharchanka:
Can't test it myself (I'm still having networking issues on WSL2), but if it works that's great.

PR #26, opened 2020-09-02
TPL build dependencies

mdavis36:
# Summary

This PR fixes the issue of building TPLs where one of the dependencies is provided by the user at a pre-installed location. 

It works by checking if a target will be built or not, setting a variable `<tpl_name>_build_dep` to either `empty` or the `<tpl_name>`. The `<tpl_name>_build_dep` variable is then referenced in the DEPENDS list of the ExternalProject_Add of the dependent target.

### Testing

I have built and tested this by providing my own pre-built and installed version of hdf5 and passing that to cmake. This built and passed all tests. However I was not able to build using the installed hdf5 in gapps/Spheral/tpl... .


jmikeowen:
I think I just fixed the zlib dependency issue.

jmikeowen (approved):
Looking good I believe.

PR #27, opened 2020-09-03
[WIP] Squashed Warnings Pyb11 Modules

mdavis36:
# Summary
This PR squashes all warnings for a complete Spheral Python modules build with GCC.

## Changes
- Set CMake warnings flag to ON.
- Silence unused-local-typedef warnings for all Spheral python modules. This is necessary due to the way in which PYB11Generator creates the pybind11 .cc Spheral files.
- Silence terminate warning on SpheralSilo.cc. This is tracked down to a VERIFY2 macro call from DBC.hh which throws this error. (An issue will be created to track this in the future).
- Silence uninitialized-variable from bisectSearch. This is being caused by some issue in the boost multi-array Iterator. (Issue will be created).

- Various unused-parameter, sign-conversion, unused-variable warnings throughout.
- Pybind11Wraps function signature fixes. Missing args, const/non-non const, return types etc.

### DIFF
--- a/cmake/SetupSpheral.cmake
+++ b/cmake/SetupSpheral.cmake
@@ -16,7 +16,7 @@ set(CMAKE_EXPORT_COMPILE_COMMANDS On)
 #-------------------------------------------------------------------------------
 # Optionally suppress compiler warnings
 #-------------------------------------------------------------------------------
-option(ENABLE_WARNINGS "show compiler warnings" OFF)
+option(ENABLE_WARNINGS "show compiler warnings" ON)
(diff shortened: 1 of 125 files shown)


jmikeowen:
I saw the commit turning off warnings for unused typedefs in PYB11 modules.  I think this is a good change -- it's actually more work for the developer adding & removing typedefs for each class I think, and having some blanket typedefs does not seem harmful.

jmikeowen (approved):
Builds and tests properly for me, so I'm good to go.

MishaZakharchanka (approved):

PR #30, opened 2020-09-21
Squashing debug mode warnings

mdavis36:
This PR squashes warnings generated in gcc during a debug build. Almost all warnings were sign conversion warnings.


jmikeowen (approved):
Looks good to me!  Clearly we're going to have to add debug & contract builds to our standard set when we add automated test builds.

PR #31, opened 2020-09-23
Clang Build Fixing Errors and Warnings

mdavis36:
# Summary
Fixing errors and warnings for building spheral with clang. Clang seems to be a little more picky so it was able to pick up some things that weren't reported with gcc.

## Changes
- Commenting out unused arguments.
- Type signage warnings.
- Swapped std -> boost for unordered_map. It appears that with C++11 the stdlibs for clang does not have support for unordered_map. I believe this change was previously made in some other places of spheral. Now all unordered_map calls are pulled from boost.
- Override declarations, clang wanted explicit override calls on virtual function implementations.
- Reverted the addition of the dt function in DamageModel.hh. This was reverted from changes made in the gcc fix as it is never implemented and created an undefined symbol when building with clang, breaking the Spheral python module.
- Mismatched struct/class declaration of DomainNode and FieldListSet.
- Removed duplicate struct definition of KeyTraits in GeomPolygon.
- Removed unused functions in polyclipper2d.cc and fillFacetVolume.cc
- -Wno-terminate only needed for gcc.
- Attempt to fix the undefined-var-template warnings. These are produced when using a static variable of a class template that is initialized in another translation unit. For the most part his was fixed e.g. GeomVector, Geom<N>RankTensor. This is done by instantiating the static variables within the headers. However I was unable to fix this with GeomTensor, This is because GeomTensor::zero is not called ever in the compiled C++ Geometry library, leaving them undefined, thus throwing an undefined symbol error when importing the Spheral python module. There are other fixes for this but I believe they are beyond the scope of this PR so I have turned off that warning for all builds in the meantime.


MishaZakharchanka (approved):

jmikeowen (approved):
My only question is about moving back to boost::unordered_map from std::unordered_map.  We've been trying to go to the std:: versions as they're available -- what's up with clang?

PR #32, opened 2020-09-28
Feature/damage

jmikeowen:
Largely a bugfix merge to date:
 - fixed a problem with time advancement when allowing time integrators to dynamically fallback due to unstable timesteps
 - improvements to filling polyhedral shapes for problem generation
 - from Brody:
   - fixes for temperature lookup in gamma law gas equation of state
   - making some test more reproducible by using fixed random number seeds


src/Material/GammaLawGas.cc
```diff
@@ -53,7 +53,7 @@ setPressure(Field<Dimension, Scalar>& Pressure,
             const Field<Dimension, Scalar>& massDensity,
             const Field<Dimension, Scalar>& specificThermalEnergy) const {
   CHECK(valid());
-  for (size_t i = 0; i != massDensity.numElements(); ++i) {
+  for (auto i = 0; i != massDensity.numElements(); ++i) {
```
mdavis36, line 56: `auto` evaluates to a signed integer here which throws a sign compatibility warning between `i` and `massDensity.numElements()` since `i` will never be negative we should keep it as an unsigned type. This is also seen at lines 71, 86 and 159 of this file

src/NodeGenerators/fillFacetedVolume.cc
```diff
@@ -202,6 +242,24 @@ fillFacetedVolume10(const Dim<3>::FacetedVolume& outerBoundary,
...
   }
+
+  // If we didn't find anything, fall back to sampling on the surface.
+  if (allReduce(result.size(), MPI_SUM, Communicator::communicator()) == 0U) {
+    if (Process::getRank() == 0) {
+      cerr << "Falling back to surface points..." << endl;
+      const auto nexpect = std::max(1, std::min(int(verts.size()), int(outerBoundary.volume()/(dx*dx*dx) + 0.5)));
+      std::vector<unsigned> iresult(verts.size());
+      for (auto i = 0U; i < verts.size(); ++i) iresult[i] = i;
+      while (iresult.size() > nexpect) {
```
mdavis36, line 253: A sign warning is thrown between `iresult.size()` and `nexpect` on 253. `nexpect` is evaluated as an integer from `std::max`, since it will never be negative could we make it unsigned?

mdavis36 (changes requested):
Just some quick comments but everything is building and testing fine for me :+1:

jmikeowen:
Boy, this type stuff is going to drive me mad!  But yep, I fixed those types in the latest commit.

src/NodeGenerators/fillFacetedVolume.cc
jmikeowen: Fixed in the latest commit.

mdavis36 (approved):

PR #34, opened 2020-10-22
Python virtual-env ; gapps TPLs ; lc-build scripts

mdavis36:
# Summary

This PR allows Spheral to be built off of entirely pre-built TPLs. This feature is being added in the hope that we can compile Spheral without needing to compile any TPLs on LC machines (most of the time). This will help with cutting down on a lot of the upfront compile time for a clean Spheral build, and also helps with subsequent builds as less target checks need to be run on each make call.

### Install Directory/ Python Virtual Environment
The proposed changes bring alterations in the way that an install directory will look for Spheral. 

- I have added a hidden python virtual environment `.venv` that resides in the install directory. The virtual environment will point to the python install the user wishes to use (either built or pre installed somewhere else) and will import all of the pre-installed pip modules. The compiled Spheral python files will be aliased into the `.venv/.../site-packages` dir from the `CMAKE_INSTALL_PREFIX/Spheral` directory. 
- `SPHERAL_INSTALL_DIR` -> `SPHERAL_TPL_DIR` : A name change for this variable as it is really only used as the base directory to **search for or install** any TPLs who's locations aren't otherwise defined by the user. The new default location for this variable is `CMAKE_INSTALL_PREFIX/tpl`.
- If the user builds any TPL with Spheral they will now live under `CMAKE_INSTALL_PREFIX/tpl`, previously they would live directly in `CMAKE_INSTALL_PREFIX`. It is stil an option for users to define specific installation / search locations for each TPL with `<tpl-name>_DIR`.

### Executable Scritps
- ATS tests can be run on LC systems with the use of the `atstest` executable, found in the install directory. This command forwards the correct python location for the build to ats so Spheral user `ats` commands will change from:
`/usr/gapps/Spheral/bin/lcats -e <python_prefix>/bin/python2.7 <spheral_src_dir>/tests/integration.ats`
to:
`./<spheral_install_dir>/atstest <spheral_src_dir>/tests/integration.ats`
The `atstest` script is useful due to addition of the python virtual environment. With these changes you would need to manually activate the virutal-environment before running ATS and then manually deactivate afterwards. This helps to streamline that process.
- The `spheral` alias in the installation directory now aliases a script that has similar functionality to the ATS script. It invokes the virtual-env, calls the correct python with forwarded command line arguments and deactivates the virtual environment after. From the user's perspective this should appear to function exactly the same as before.

### Gapps TPLs
- Within the Spheral gapps space a few common TPL configurations have been installed. This should allow users to easily run fresh builds of Spheral without rebuilding TPLs each time. The TPLs can be easily pointed to for a build by including a host-config file in your `cmake` command. e.g:
`cmake <spheral_src_dir> -C <spheral_source_DIR>/host-configs/.../gccX_tpl.cmake`
These host configs can also be edited easily to point to other installed TPLs you are commonly using i.e. specific HDF5 builds.
- Currently there only exist release tpl builds for TOSS systems with gcc and clang.

### LC-scripts
- A variety of LC scripts have been added for TOSS systems. Each script builds out a small directory tree under the name of the script to compartmentalize build configurations. With the addition of these scripts and the gapps TPL all that is required for a fresh build of a given configuration is to execute one of the scripts from the spheral root directory:
`./scripts/lc-builds/toss3_gcc-8.3.1-release-mpi-python.sh`
`cd lc_toss3-gcc-8.3.1-rel-mpi-py/build`
`make -j install`
`../install/atstest ../../tests/integrations.ats`


jmikeowen:
Just tried to build this branch on my WSL2 machine, but maybe there's a file missing:

-- Installing: /home/owen/projects/Spheral/optimize/scripts/spheral-env.sh
-- Installing: /home/owen/projects/Spheral/optimize/scripts/atstest.sh
Creating Spheral virtual python environment ...
./scripts/spheral-setup-venv.sh: line 3: virtualenv: command not found
Installing Spheral ...
cp: cannot create regular file '/home/owen/projects/Spheral/optimize/.venv/lib/python2.7/site-packages/': No such file or directory
cp: cannot create regular file '.venv/lib/python2.7/site-packages/': No such file or directory
Installing Gnuplot, ats ...
Creating spheral symlink to spheral-env script ...
make: Leaving directory '/home/owen/projects/Spheral/optimize/BUILD'

mdavis36:
Looks like you don't have the virtualenv utility available. I'll add it as a python module with our pip packages then we can use a local one.

jmikeowen:
One step closer!  It installs the virtual environment, but it doesn't work from csh.  Seems fine from bash.  I know virtual env works for either shell, so gotta figure out what's busted with csh.

mdavis36:
To use alternative TPLs with the lc-build scripts you can append overriding CMake arguments to the end of each script call 

To **SEARCH** for an existing install:
`./scripts/lc-builds/toss3_gcc8.3.1-release-mpi-python.sh -Dboost_DIR=<full-path-to-boost-install>`

To **BUILD** a local version of the TPL:
`./scripts/lc-builds/toss3_gcc8.3.1-release-mpi-python.sh -Dboost_BUILD=On -Dboost_DIR=""`
or:
`./scripts/lc-builds/toss3_gcc8.3.1-release-mpi-python.sh -Dboost_BUILD=On -Dboost_DIR=<local-dir-to-install>`

jmikeowen:
Question about local individual third party libs.  Can we check if the locally specified lib (say Boost for instance) has a different version than the prebuilt one, and if so use that one?  By far the most common use case I have for this would be I'm testing updating a library, and so it'd be cool to just update the appropriate cmake TPL lib spec and have it use that one automatically.  This is how we handled it in Kull as I recall.

mdavis36:
The way the lc-scripts work is by setting the boost_DIR variable in a host-config cmake file (`host-configs/`). With the extended arguments above all we are doing is overwriting it manually for the CMake configuration command. The `*_DIR` variables are the only directory associated with a given TPL, the behavior of the TPLs are handled by other flags e.g. `boost_BUILD` and `BUILD_TPLS`. 

I think what you want is possible, however it will require a few design changes to the build system, namely the ability to define a pre-installed and a local `*_DIR` variable for each TPL. It would also require that we write custom comparative scripts for each TPL as some TPLs reveal their version information through a `.settings` file in their libs dir, others we would have to extract from other places such as header files. For these reasons if we really want this feature I think it might be appropriate to tackle it in a separate PR.

In the meantime If you would like all of your builds to point to a specific TPL dir locally you can do that by temporarily editing the host-config file e.g. `host-configs/lc-builds/toss3/gccX_tpl.cmake`. This will save you from having to redefine the custom TPL location each time you run one of the scripts.

jmikeowen:
Ah, well then I think the behaviour I'm looking for is outside the scope of this PR then.  It might be worth looking at for a future change focused on this sort of feature, but let's not feature creep on this one.

jmikeowen:
Another suggestion -- the canned build scripts for LC are nice, but they are hardwired assuming you're building in a subdirectory of the source tree.  I never do this, as I always try to keep the build generated files entirely outside the git controlled source.  How about an optional argument to those scripts indicating the path to the source?  This can default to ../.. to reproduce the current behaviour, but let us build elsewhere if so desired.

mdavis36:
Sure I'll add that, I was thinking about that when editing the build docs today. Personally I keep them in the source tree as I like to have a couple of different spheral source directories, then I don't get all of my build directories mixed up.

mdavis36:
I've added a -d name argument to the lc build scripts. This should allow us to generate the build directories anywhere. You can pass either the relative or full path to the spheral root directory.

`<spheral_src>/scripts/lc-builds/toss3_gcc8.3.1-release-mpi-python.sh -d <spheral_src_path> -D<CMAKE_ARGS1> -D<CMAKE_ARG2>`

jmikeowen:
Another question about the lc scripts -- the install path is also hardwired, which is not convenient when we want to use these scripts to install releases for public use.  I'd suggest making the install path something that can be specified as well, keeping the default the current behaviour.

jmikeowen:
Looks good on LC, including using tcsh.

jmikeowen (approved):
I think we're good to go with this PR now.  There may be more testing needed on Ubuntu and WSL2 to come.

PR #39, opened 2020-12-04
Fixing conduit and axom external project src directory.

mdavis36:
These changes should help with the conduit configure errors some people were seeing.

### DIFF
--- a/cmake/tpl/axom.cmake
+++ b/cmake/tpl/axom.cmake
@@ -36,7 +36,7 @@ if(${lib_name}_BUILD)
                -DENABLE_TESTS=Off
 
                -DCONDUIT_DIR=${CONDUIT_INSTALL_DIR}
-               -DHDF5_DIR=${hfd5_DIR}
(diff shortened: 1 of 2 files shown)


jmikeowen (approved):
This looks safe enough.  So has anyone had a chance to test this with the older CMake yet?  In particular the 3.10 that comes with Ubuntu was failing before.

mdavis36:
I tried it with 3.12 as that is the closest on LC. The 3.10 documentation shows the SOURCE_SUBDIR command is available (so this *should* work as expected). I can give this a go later in a WSL2 build but I don't have that environment set up on any of my systems, so it may be a little.

jmikeowen:
Yeah, I actually tried the old way with CMake 3.12 on LC and it worked too.  I'll try to test a standard Ubuntu install in WSL2 as well and see if the default 3.10 works now.  If so I'll amend the build instructions.

MishaZakharchanka (approved):
These changes work on my end.

Sorry for getting to this late, didn't get a notification for some reason.

PR #40, opened 2020-12-21
Fixing install scripts w/ srun on LC.

mdavis36:
Slurm did not know which shell to use with some of our installed scripts.


jmikeowen (approved):
Yep, this fixes the srun issue!

PR #41, opened 2021-01-04
Named timing output file

mdavis36:
This adds the ability to define a name for a given output timing file. If no name is passed to the SpheralController the default  name`time.table` will be used and behavior will remain the same.

### DIFF
--- a/src/Utilities/Timer.hh
+++ b/src/Utilities/Timer.hh
@@ -64,10 +64,7 @@ public:
   static std::list<Timer*> TimerList;
 
   static void TimerSummary(const std::string& fname);
-  static void TimerSummary(){ TimerSummary("time.table"); }
-  static void TimerSummary(const int, const int) {
-    TimerSummary(); // backwards compatibilty...
-  }
+  static void TimerSummary() { TimerSummary("time.table"); }
   
 private:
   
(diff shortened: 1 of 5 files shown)


src/Utilities/Timer.hh
```diff
@@ -63,12 +63,12 @@ public:
...
   static std::list<Timer*> TimerList;
 
-  static void TimerSummary(const int bert, const int ernie) {
+  static void TimerSummary(const std::string& fname);
+  static void TimerSummary(){ TimerSummary("time.table"); }
+  static void TimerSummary(const int, const int) {
     TimerSummary(); // backwards compatibilty...
   }
   
-  static void TimerSummary(void);
```
mdavis36, line 70: @jmikeowen I know passing void as an argument type is a C trick for explicitly defining a function takes 0 arguments. Does removing this break some interoperability with the C wrapping library that I may not be aware of?

jmikeowen: I don't think you're breaking anything there.  This Timer class stuff is a really old version of something constructed by Noel Keen years ago, who now works at Lawrence Berkeley Lab still on performance stuff.  You could ask him (NDKeen@lbl.gov) for more info on why he did it this way, and he likely has some updated version of this by now.

jmikeowen (approved):
All looks reasonable to me.

PR #44, opened 2021-01-08
Removing expensive timers from SPHEvaluateDerivatives.

mdavis36:
Removes timer code from SPHEvaluateDerivatives and SolidSPHEvaluateDerivatives which were causing performance hits.


jmikeowen (approved):
I added some more instances of the same pattern as well, and tested it out on LC.   Looks good.

PR #45, opened 2021-01-13
GCC 8.1 lc-scripts for tpls and build; R-Path fix for TPL local Spheral builds; Script directory name changes.

mdavis36:
This PR does a few things:
- Adds two new lc-build scripts for gcc8.1.0. This PR adds a tpl-only script and a gcc8.1.0 release Spheral build for python with MPI support. 
- Changes to host-configs. host-configs are now compiler version dependent.
- Fixes an issue where installed Spheral libs lose track of their python and boost linked libs. This only happens when building the TPLs within the Spheral build itself.
- Scripts generate directories based on the name of the script itself. This is useful for keeping things consistent and for some future CI functionality that will rely on utilizing lc-scripts and their tree structures.
- A few warnings and other fixes found on the way.


jmikeowen (approved):
Looks good -- thanks for getting this fixed!

PR #48, opened 2021-01-20
Update building.rst

as1m0n:
1) Changed "chdir" to more universal "cd"
2) changed ../../spheral to ../../Spheral to match current setup

### DIFF
--- a/docs/building.rst
+++ b/docs/building.rst
@@ -27,7 +27,7 @@ I like to keep my build & install files separate from the git cloned source, so 
 
   git clone --recursive https://github.com/jmikeowen/Spheral
   mkdir -p Spheral_release/BUILD && cd Spheral_release/BUILD
-  cmake -DCMAKE_INSTALL_PREFIX=`chdir ..; pwd` ../../spheral
+  cmake -DCMAKE_INSTALL_PREFIX=`cd ..; pwd` ../../Spheral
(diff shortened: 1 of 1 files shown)


jmikeowen:
The "cd" change seems reasonable (this is shell dependent, but probably more folks are using bash these days).  But the default clone of Spheral drops in a directory called "spheral" with the lower-case s doesn't it?

as1m0n:
Mike,

On my system (Windows 10 with WSL2, Ubuntu 20.04) there are two folders, 
Spheral (after git clone etc...) and Spheral_Release (after the script 
mkdir, etc.).
Is there a 3rd folder "spheral" that should be made? There is no 
"../../spheral".

Andy

------ Original Message ------

as1m0n:
I ran the "git clone --recursive https://github.com/jmikeowen/Spheral" 
and it places everything into "Spheral".

Andy

------ Original Message ------

jmikeowen:
Ah, I see the difference.   If you use the clone command that the GitHub site offers up on the "Download" button, it comes out lower-case.  But, because I wrote the upper-case Spheral in the docs you're getting it that way.  This makes your change to the docs more consistent, so we should stick with that.

jmikeowen (approved):
This is more consistent with bash, which is a common shell, so looks good.

PR #52, opened 2021-02-18
Feature/expanding domain

jmpearl:
Class that takes a user defined python function and uses it to feed nodes into the simulation. Currently restricted to ellipsoids/spheres for the initial sub-region. Tested for RZ/2d/3d, fluid/solid, & multiple nodelists


jmikeowen:
This looks pretty reasonable.  My only request is that we move the class to the src/Boundary package, since this is really a boundary condition.

jmpearl:
sounds good, I'll shift it over

jmikeowen (approved):
All looks good now.

PR #55, opened 2021-02-27
Feature/spherical sph

jmikeowen:


jmikeowen:
This merge brings up infrastructure developed to support the spherical (and eventually cylindrical) symmetry kernel formalism.  The spherical kernel is pretty much working now, but the corresponding spherical SPH is not yet.  I'm bringing this branch up to develop now because the infrastructure developed here (particularly the quadratic and bi-quadratic interpolators) are more generally useful (and needed for other work such as improving our ANEOS implementation.

This merge also brings updates to 3 third party libs (boost, Py11Generator, and some PIP modules).  I've confirmed these are all working building stand alone, as well as updated the pre-installed TPLs for the GCC versions on LC machines.

MishaZakharchanka (approved):

src/Kernel/Kernel.hh
```diff
@@ -86,13 +75,13 @@ public:
   // All Kernels are required to define the "kernelValue", "gradValue",
   // and "grad2Value" methods, with the same call signatures 
   // as these functions.
-  double kernelValue(double etaMagnitude, double Hdet) const;
-  double gradValue(double etaMagnitude, double Hdet) const;
-  double grad2Value(double etaMagnitude, double Hdet) const;
+  double kernelValue(double etaMagnitude, const double Hdet) const;
```
brbass, line 78: Is there some reason why Hdet is const and not etaMagnitude? Is etaMagnitude changed in some of the routines?

jmikeowen: Offhand I think no particular reason.  We could make the etaMagnitude constant as well.  Is that a change you'd like to see?

src/Geometry/GeometryRegistrar.hh
```diff
@@ -0,0 +1,45 @@
+//------------------------------------------------------------------------------
+// GeometryRegistrar
+//
+// A singleton class for maintaining global information about the type of
```
brbass, line 4: Might be worth considering certain things on Geometry rather than Dimension later on.  That might cut down on duplicate code between Cartesian/Cylindrical/Spherical geometry as you could specialize things like kernel normalization, integration, etc. while keeping a lot of the machinery the same between them.

src/Kernel/Kernel.hh
brbass: It doesn't matter that it happens now, but it would be nice to keep things consistent between TableKernel, SphericalTableKernel, and Kernel.

PR #57, opened 2021-03-09
Enable polyclipper shared tpl

mdavis36:
This PR enables the use of the built PolyClipper TPL in the shared usr space of LC.
Updates some of the TPL only install scripts for use on the LC.
Adds a couple of blueos scripts that I am using for building with raja/cuda.


jmikeowen:
Overall looks good, only one problem.  The version of PolyClipper Spheral now expects is 1.12, and we have 1.11 installed in the shared LC area.  We should update the installed TPL PolyClipper before pushing this change up.

jmikeowen (approved):
Looks good to go!

PR #58, opened 2021-03-10
Switched StrainPorosity velocity gradient

jmpearl:
Changed the velocity gradient that StrainPorosity uses to calculate distension time-derivative from the nodelist-local velocity gradient to the nodelist-global version. This should give better consistency w/ the density time-derivative which also uses the global version. The difference seemed to cause some ringing in the distension/density for quasi-steady state interactions of different nodelists on longer run-outs.


jmikeowen (approved):
Looks good.  This should be a very subtle change for most applications.

PR #59, opened 2021-03-10
Bugfix Intel Build

mdavis36:
# Summary
This PR is a working bug fix for performing a full C++ & Python Spheral build with Intel. Currently there are existing build issues at the Python library build stage.

This PR adds: 
- Compilers.cmake file for defining compiler specific cmake variables and flags.
- Fix to building `OpenSubDiv` on LC with Intel compilers.
- Adds some Intel build scripts.

## Notes:
- Performing `make -j` on the build seems to error out "randomly" perhaps we are running out of memory during a build with intel? Using `make -j 16` seems to work fine.
- This PR is for fixing Spheral Python builds the C++ only build compiles fine.


jmikeowen:
I know this is still a draft, but I have noticed in the past we need to build with no more tasks than available processors (so "make -j N", where N < number of available cores), and since Intel really pushes the memory during it's build I'm not surprised it fails with just -j.  Is it ever a good idea to fire off more simultaneous build tasks than available resources?

mdavis36:
I`m not aware that it will ever attribute to a performance hit in the build, but I can't be certain. I would think it only matters when you hit a memory limit like I believe is happening with the Intel build. I know when building the gcc compiler locally their user guide actually recommends using more processes than cores if possible to speed up compilation, so ... :man_shrugging: . I only have the defined -j16 in the build scripts for now in the case that this doesn't get touched for a little while and we forget about it. Ultimately I would be inclined to leave it up to the user to define how much parallelism they want to use in their Spheral build

mdavis36:
@jmikeowen TPLs are updated on LC and all tests are passing on a build.

Some calls to `FlatFileIO2d` exist in the following python files:
`src/ExternalForce/tests/TestNFWPotential.py`
`src/ExternalForce/tests/TestPointPotential.py`
`src/SimulationControl/findLastRestart.py`

MishaZakharchanka (dismissed):

PR #62, opened 2021-03-25
TPL hash install dirs

mdavis36:
This PR installs TPLs under a directory tagged with their download hash. This helps differentiate TPL installs in the shared gapps space allowing multiple versions of TPLs to maintain backwards compatibility.

This PR also updates the PolyClipper TPL to a version with a clang build fix. 

The TPLs have been updated in the shared space


jmikeowen (approved):
Looks good to me.

PR #63, opened 2021-04-15
Update SynchronousRK2.cc

jmpearl:
Flipped ordering of initializeDerivatives & derivs.zero to be consistent w/ the other integrators.


jmikeowen:
Ah, makes sense to be consistent.  Have you run the full ATS test suite to ensure everything is passing with this change?

jmpearl:
ATS came back clean

PR #64, opened 2021-04-20
Feature/sidre data collection

MishaZakharchanka:
Adding functions to store DataTypes into sidre and tests for the those functions.


src/CXXTests/testSidreStorage.cc
```diff
@@ -0,0 +1,645 @@
...
+  Spheral::Field<Spheral::Dim<1>, Spheral::Dim<1>::Vector> testField("test field", makeNodeList);
+  for (int i = 0; i < n; i++)
+    testField[i] = Spheral::Dim<1>::Vector(i);
+  
+  double* rawSidreData = myData.alloc_view("SidreTest", testField)->getData();
+
+  //myData.printDataStore();
+
+  //for (int i = 0; i < n; i++)
+    EXPECT_EQ(testField[0].x(), rawSidreData[0]);
```
mdavis36, line 190: We should ensure we are checking every element of the sidre data for each of these tests.

src/CXXTests/testSidreStorage.cc
```diff
@@ -0,0 +1,645 @@
...
+//   testField[0] = Spheral::Dim<1>::FacetedVolume(temp);
+    
+  
+//   double* rawSidreData = myData.alloc_view("SidreTest", testField)->getData();
+
+//   myData.printDataStore();
+
+//   // for (int i = 0; i < n; i++)
+//   //   EXPECT_EQ(testField[0](0, 0, 0, 0, 0), rawSidreData[0]);
+// }
```
mdavis36, line 341: Please make a pass through all of the touched files to remove any old / unused or commented out code like this.

src/CXXTests/testSidreStorage.cc
```diff
@@ -0,0 +1,645 @@
...
+
+  int result = RUN_ALL_TESTS();
+
+// #ifdef AXOM_USE_MPI
+//   MPI_Finalize();
+// #endif
+
+  return result;
+}
+
```
mdavis36, line 645: I don't think this is needed for a gtest file. CMake and googletest will make their own executable which wraps the above tests.

src/Pybind11Wraps/Sidre/Group.py
```diff
@@ -0,0 +1,36 @@
...
+    #     "Return pointer to non-const View with given name or path."
+    #     return "axom::sidre::View*"
+
+    # @PYB11namespace("axom::sidre::View")
+    # def getData():
+    #     "Return data held by view and cast it to any compatible type allowed by Conduit (return type depends on type caller assigns it to)."
+    #     return "axom::sidre::Node::Value"
+
+    #...........................................................................
+    # Attributes
```
mdavis36, line 36: If this file is unused please remove it

src/Utilities/SidreDataCollectionInline.hh
```diff
@@ -0,0 +1,531 @@
...
+   int view_count = 0;
+   for (u_int i = 0; i < field.size(); i++)
+   {
+      auto *data = &(*field[i].begin());
+      m_datastore_ptr->getRoot()->createView(view_name + std::to_string(view_count), dtype, 
+                                                         num_elements, (void*)data);
+      view_count++;
+   }
+   return m_datastore_ptr->getRoot()->getView(view_name + "0");
+}
```
mdavis36, line 510: The majority of the alloc_view methods are pretty much identical and can be templated together, the only difference I see in most of them is the value of num_elements which can be extracted from the Fields DataType type trait. Templating this will greatly reduce code duplication here

src/Utilities/SidreDataCollectionInline.hh
```diff
@@ -0,0 +1,531 @@
+#include "axom/sidre.hpp"
```
mdavis36, line 1: We should add Comment headers to each of the new files explaining a little about what each one does

mdavis36 (changes requested):
I've left some comments for some possible refactoring and cleanup, but overall I think this is forming up well :+1:

src/CXXTests/testSidreStorage.cc
MishaZakharchanka: I do intend to do this, I just don't know the best way to do this right now as these functions aren't used anywhere. Once I start working on the restart I'll know what the functions should return and I can adjust the tests then.

mdavis36: I think you could iterate over the number of elements and just ensure that all the data in rawSidreData is as expected. In this case you should be able to do something like:
```
for (int i = 0; i < n; i++)
    EXPECT_EQ(i, rawSidreData[i])
```
for the 3D Vector below you could do:
```
for (int i = 0; i < n; i++)
    EXPECT_EQ(i, rawSidreData[i*3])
    EXPECT_EQ(i+1, rawSidreData[i*3+1])
    EXPECT_EQ(i+2, rawSidreData[i*3+2])
```

MishaZakharchanka: The way that I am storing the dim types in sidre now is data type per view (a field of dim<1>vectors has multiple views for each vector value) so, rawSidreData only has the first value. I did this because it was the easiest to make it work at the time, but it will change once I start using the functions.

jmikeowen:
So if I want to exercise the tests of the new stuff, what do I do?  Does "make test" cover it?

mdavis36:
> So if I want to exercise the tests of the new stuff, what do I do? Does "make test" cover it?

From the build dir make test should work

src/CXXTests/testSidreStorage.cc
```diff
@@ -0,0 +1,475 @@
...
+      return testField;
+    }
+
+    void allocRawSidreData(const Spheral::Field<Spheral::Dim<1>, T>& testField)
+    {
+      rawSidreData = myData.alloc_view("SidreTest", testField)->getData();
+    }
+};
+
+using MyTypes = ::testing::Types<char, int, size_t, uint32_t, uint64_t, float, double>;
```
mdavis36, line 32: Lets rename this type alias to something more descriptive

src/CXXTests/testSidreStorage.cc
```diff
@@ -0,0 +1,475 @@
...
+      for (int i = 0; i < n; i++)
+        testField[i] = i;
+      return testField;
+    }
+
+    void allocRawSidreData(const Spheral::Field<Spheral::Dim<1>, T>& testField)
+    {
+      rawSidreData = myData.alloc_view("SidreTest", testField)->getData();
+    }
+};
```
mdavis36, line 30: Let's move all helper classes, functions, includes and type aliases into a separate header file keeping only the gtest "TEST" type calls in this file.

src/Utilities/SidreDataCollectionInline.hh
```diff
@@ -0,0 +1,145 @@
...
+      auto *data = &(*field[i].begin());
+      m_datastore_ptr->getRoot()->createView(view_name + std::to_string(view_count), dtype, 
+                                                         num_elements, (void*)data);
+      view_count++;
+   }
+   return m_datastore_ptr->getRoot()->getView(view_name + "0");
+}
+
+
+}
```
mdavis36, line 145: Closing out namespace scopes with a comment can be really helpful in the future 
```
} // namespace Spheral
```

src/CXXTests/CMakeLists.txt
```diff
@@ -3,26 +3,25 @@
...
 #------------------------------------------------------------------------------
 # Add gtest C++ tests
 #------------------------------------------------------------------------------
+
+get_property(SPHERAL_CXX_LIBS GLOBAL PROPERTY SPHERAL_CXX_LIBS)
+
 foreach(test ${gtest_spheral_tests})
     get_filename_component( test_name ${test} NAME_WE )
-    blt_add_executable( NAME ${test_name}_test
+        blt_add_executable( NAME ${test_name}_test
```
mdavis36, line 18: Formatting

mdavis36 (changes requested):

jmikeowen:
In src/Utilities/SidreDataCollection, wouldn't it be preferable to use a std::shared_ptr<axom::sidre::DataStore> rather than a bare C pointer?  This way you can construct using std::make_shared, and not worry about deallocating in the destructor.  In general I try to avoid using bare pointers with new/delete just to avoid memory confusion down the line.

src/Utilities/SidreDataCollectionInline.hh
```diff
@@ -0,0 +1,297 @@
...
+    template<typename Dimension>
+    axom::sidre::Group *sidreStoreField(const std::string &view_name,
+                                  const Spheral::Field<Dimension, Dim<2>::Vector> &field);
+
+        
+private:
+    std::shared_ptr<axom::sidre::DataStore> m_datastore_ptr = std::make_shared<axom::sidre::DataStore>();
+};
+
+} // namespace Spheral
```
mdavis36, line 66: Why is this class declaration duplicated here?

src/Utilities/SidreDataCollectionInline.hh
```diff
@@ -0,0 +1,297 @@
...
+      view_count++;
+   }
+   return wholeField;
+}
+
+//------------------------------------------------------------------------------
+template<typename Dimension>
+inline
+axom::sidre::Group *SidreDataCollection::sidreStoreField(const std::string &view_name, 
+                                                   const Spheral::Field<Dimension, std::string> &field)
```
mdavis36, line 106: Formatting across this file

src/Utilities/SidreDataCollection.hh
```diff
@@ -0,0 +1,70 @@
...
+              typename std::enable_if<std::is_arithmetic<DataType>::value,
+                                      DataType>::type* = nullptr>
+    axom::sidre::Group *sidreStoreField(const std::string &view_name,
+                                  const Spheral::Field<Dimension, DataType> &field);
+
+    template <typename Dimension, typename DataType,
+              typename std::enable_if<is_rank_n_tensor<DataType>::value && !std::is_arithmetic<DataType>::value,
+                                      DataType>::type* = nullptr>
+    axom::sidre::Group *sidreStoreField(const std::string &view_name,
+                                  const Spheral::Field<Dimension, DataType> &field);
```
mdavis36, line 33: Formatting

mdavis36 (changes requested):

src/Utilities/SidreDataCollectionInline.hh
```diff
@@ -0,0 +1,214 @@
...
+      view_count++;
+   }
+   return wholeField;
+}
+
+//------------------------------------------------------------------------------
+template<typename Dimension, typename DataType>
+inline
+axom::sidre::Group *SidreDataCollection::sidreStoreField(const std::string &view_name,
+                                                         const Spheral::Field<Dimension, std::tuple<DataType, DataType, DataType, DataType, DataType>> &field)
```
MishaZakharchanka, line 117: The parameters are lined up, but the second one is very long. Do I need break it up onto more lines?

mdavis36: It's not the prettiest, if you can break it up into a format that maintains the readability go for it, but I don't think it's too much of an issue, this only really effects the tuple cases.

src/CXXTests/CMakeLists.txt
mdavis36: Bump

mdavis36:
@jmikeowen I think we might have the "make test" ats test added in a separate PR as that is going to require us to do some build system trickery to encode the build directory path into the python test file and I'd like to talk about some ideas.

@MishaZakharchanka Please fix the formatting in src/CXXTests/CMakeLists.txt. After that I'll be happy to approve this PR. 
LGTM :+1:

jmikeowen:
Sure, we can talk about how to add the CXX tests.  I'm gonna be traveling
until next week, but we can do this sometime then.

MishaZakharchanka:
As of now this branch is up to date with develop, so we should be good to merge.

mdavis36 (approved):
LGTM

jmikeowen (approved):
Looks good to go.

PR #65, opened 2021-04-21
Added a circle-3d line-2d clipping feature for inflow

jmpearl:
Made a change to the finalize method to allow only a subset of the nodes to be converted from ghosts to internals. This allows the radius of the inflow "beam" to be adaptively changed throughout the simulation.


jmikeowen:
I'm not sure I quite understand here -- doesn't this break the ordinary inflow/outflow over planes expected default behavior?  It looks like these circle parameters default such that points that cross the inflow/outflow planes will fail their tests unless the user also set's something for this "beam" parameter.

jmpearl:
Ooo you're right, I overlooked that. I'll go back in there and get the default behavior consistent.

jmpearl:
Hey Mike, the clipping mod is backwards compatible now. I clean up the implementation a bit -- what I had before was a little convoluted.

PR #66, opened 2021-04-26
Bugfix/build without network

jmikeowen:


jmikeowen:
This merge updates our TPL build rules so we can successfully build when there is no network connection.  This relies on either pre-installing the TPLs, or providing all the tarballs and such in the build tree tpl/cache directory.

Tested builds with and without network, as well as with and without the cached TPLs when the network is present.  Also tested all regression tests.

mdavis36 (approved):
Sorry it took a while to get around to this, this PR looks good to me :+1:

MishaZakharchanka (approved):

PR #67, opened 2021-05-26
Water EOS

jmpearl:
Added the stiffened gas equation of state to Material. 

Looked like there was a typo in MurnaghanEOS. Moved bulk mod K to the numerator for P = K/n * (eta-1)^n


jmikeowen (approved):
I must have been a bit sloppy putting this one in, so looks good.

PR #68, opened 2021-05-27
Feature/variable resolution damage

jmikeowen:
This change is focused on updates to our damage models to make them more efficient and able to properly handle damage in the presence of varying initial resolution.  There are two major algorithmic changes:

1. GradyKippTensorDamageOwen was an existing algorithm to handle this situation, but was way to slow during problem generation.  I've implemented support for per point random number generators, and used this to more effectively parallelize problem setup of this model.  It is now orders of magnitude faster in typical problems (minutes to setup rather than hours).  However, this model still requires explicitly holding all the activation flaw strains, which for hi-res impact problems can be many billions and overflow memory on achievable machine allocations.

2. We have a new model (ProbabilisticTensorDamage) which implements a similar algorithm to GradyKippTensorDamageOwen, but based on statistics is able to represent the span of flaws on each point mathematically.  This reduces our storage per point to a few doubles and some relatively quick math to update the damage when the implicit flaw distribution is activated.  The results are similar to the older algorithm, run just as quickly, but require very little memory.

I have added regression tests of the new model in TensileRod-1d, and tested it out for 1d and 2d tensile rods, as well as hypervelocity cratering impact models.

The bulk of the files touched in this merge are just to support Fields of std::uniform_distribution random number generators.


MishaZakharchanka (approved):

src/Damage/ProbabilisticDamagePolicy.cc
```diff
@@ -0,0 +1,346 @@
...
+//------------------------------------------------------------------------------
+// Equivalence operator.
+//------------------------------------------------------------------------------
+template<typename Dimension>
+bool
+ProbabilisticDamagePolicy<Dimension>::
+operator==(const UpdatePolicyBase<Dimension>& rhs) const {
+
+  // We're only equal if the other guy is also an increment operator.
+  const ProbabilisticDamagePolicy<Dimension>* rhsPtr = dynamic_cast<const ProbabilisticDamagePolicy<Dimension>*>(&rhs);
```
mdavis36, line 337: It look like this just performing an equivalence operation on type compatibility correct? Do we have type traits for Policies like these to make these kinds of comparisons, not a hard PR change requirement but if we make these types of comparisons for other policies that could be something to think about adding later on

src/Damage/ProbabilisticDamagePolicy.cc
```diff
@@ -0,0 +1,346 @@
...
+  if (rhsPtr == 0) {
```
mdavis36, line 338: Can we make this `nullptr` please

src/FileIO/SiloFileIO.cc
```diff
@@ -571,6 +623,17 @@ SiloFileIO::read(unsigned& value, const string pathName) const {
...
 }
 
+//------------------------------------------------------------------------------
+// Read a size_t from the file.
+//------------------------------------------------------------------------------
+void
+SiloFileIO::read(size_t& value, const string pathName) const {
+  const string varname = setdir(mFilePtr, pathName);
+  // VERIFY2(DBReadVar(mFilePtr, varname.c_str(), &value) == 0,
+  //         "SiloFileIO ERROR: unable to read variable " << pathName);
```
mdavis36, line 633: Comments

src/SimulationControl/SpheralVoronoiSiloDump.py
```diff
@@ -290,6 +290,8 @@ def dump(self, simulationTime, cycle,
 
         # Figure out how many of each type of field we're dumping.
         intFields = [x for x in self._fields if isinstance(x, eval("IntField%s" % self.dimension))]
+                     #[x for x in self._fields if isinstance(x, eval("UnsignedField%s" % self.dimension))] +
+                     #[x for x in self._fields if isinstance(x, eval("ULLField%s" % self.dimension))])
```
mdavis36, line 294: Comments

src/Utilities/uniform_random.cc
```diff
@@ -0,0 +1,147 @@
...
+}
+
+//------------------------------------------------------------------------------
+// Advance the state number of steps, as though we generated that many random
+// values.
+//------------------------------------------------------------------------------
+void
+uniform_random::advance(const size_t n) {
+  // mGen.discard(n * mGen.state_size);
+  // mNumCalls += n;
```
mdavis36, line 117: Comments

mdavis36 (changes requested):
Overall looks good, I only have a small change request for the `nullptr` otherwise I noticed some comment blocks that I'm not sure were purposely left behind along with a point of discussion on policy traits.

The random generator looks good to me, this will certainly be something that might be a tricky point for GPU implementation in the future but we can approach that when we get there, I know there are specific libraries dedicated to random number generation on the GPU we will be able to utilize.

src/Damage/ProbabilisticDamageModel.cc
```diff
@@ -0,0 +1,421 @@
...
+  // const size_t nodeListi = distance(nodeLists.begin(), find(nodeLists.begin(), nodeLists.end(), &(this->nodeList())));
+  // for (typename ConnectivityMap<Dimension>::const_iterator iItr = connectivityMap.begin(nodeListi);
+  //      iItr != connectivityMap.end(nodeListi);
+  //      ++iItr) {
+  //   const int i = *iItr;
+  //   const double D0 = damage(i).Trace() / Dimension::nDim;
+  //   dt = min(dt, 0.8*max(D0, 1.0 - D0)/
+  //            std::sqrt(mDdamageDt(i)*mDdamageDt(i) + 1.0e-20));
+  // }
+  // return TimeStepType(dt, "Rate of damage change");
```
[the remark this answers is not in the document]
jmikeowen: That is the same code from the older model (commented out in the same way).  It's not clear if we need to apply this timestep criteria or not.  Fow now I've left it there commented out and don't have the damage model voting on a timestep (it effectively depends on the hydro timestep to keep things appropriate).  I'm inclined to leave this commented out code there for future reference if we go back to having the damage need to actively vote.

src/Damage/ProbabilisticDamagePolicy.cc
```diff
@@ -0,0 +1,346 @@
...
+                                                     1.0e-5,
+                                                     1.0);
+          Di.rotationalTransform(Deigen.eigenVectors);
+        }
+        ENSURE(Di.eigenValues().minElement() >= 0.0 and
+               fuzzyLessThanOrEqual(Di.eigenValues().maxElement(), 1.0, 1.0e-5));
+        //     ENSURE(fuzzyGreaterThanOrEqual(Di.eigenValues().minElement(), 0.0, 1.0e-5) &&
+        //            fuzzyLessThanOrEqual(Di.eigenValues().maxElement(), 1.0, 1.0e-5));
+        //     ENSURE(fuzzyGreaterThanOrEqual(Di.eigenValues().minElement(), 0.0) &&
+        //            fuzzyLessThanOrEqual(Di.Trace(), 1.0));
```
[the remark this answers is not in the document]
jmikeowen: Those contracts express the limits we expect, but they're tricky to have active due to the difficulties of enforcing limits on eigenvalues of tensors (which is what these are).  I spent some time a while ago fiddling with the numerical fuzziness of the comparisons trying to make it robust, but gave up.  i also think thse are appropriate to keep at least commented out for now -- they limits described are formally correct, and if computer math was perfect we'd be good to go.

src/Damage/ProbabilisticDamagePolicy.cc
jmikeowen: That's common to many of these policices.  Might be worth a focused discussion at some point.

src/Damage/ProbabilisticDamagePolicy.cc
jmikeowen: Done in latest commit.

mdavis36 (approved):

PR #77, opened 2021-07-21
Fsisph

jmpearl:
SPH multimaterial module, here's a quick summary of the contents:
   1) SolidFSISPHHydroBase
         - solid hydro for problems with material property discontinuities
         - selectively apply density sum to individual nodelists
         - "consistent diffusion" option applies diffusion to eps and rho through the velocity gradient
         -  diffusion options for specific thermal energy and density
         -  modified damage implementation (material property reconstruction applied to damage front)
         - *currently no RZ or fluidnodelists hydro classes*
   2) SlideSurfaces
        - slide surface physics package (similar to AV implementation)
        - used to reduce artificial viscosity for slip interface conditions
        - based on interface normals between nodelists
   3) compatible energy treatment 
        - pairwise thermal energy derivatives used as weights
        - operates on the difference between the conservative and calculated total pairwise work
        - allows pairwise therm energy derivatives to have opposite signs


jmpearl:
Added Noh and Sod ats tests

jmikeowen:
This is looking pretty good -- the main thing I'd like to see before we pull is an ATS test that has pass criteria to verify FSISPH is working correctly.  Adding test conditions to the Noh/Sod tests is one possibility, but it'd be nice to have one of the test cases that FSISPH was designed for as well.

jmikeowen:
Looks good -- I'm going to approve and merge this.  I did find one testing problem -- the DiametralCompression.py test had an error in the ATS prescription.   I fixed that and pushed the change, so we should be OK to go.

jmikeowen (approved):
This is ready.

PR #80, opened 2021-08-03
Bugfix to return exit status codes from Spheral process.

jmikeowen:
When we went to using Python virtual environments to encapsulate the Spheral python, we inadvertently hid any exit codes from the actual Spheral process.  This is why our unit tests are always passing -- the unit test framework correctly detects errors and exits, but that exit code is hidden from the user by our "spheral" bash script.

This bugfix captures and returns the spheral process exit code, but I did this in a simple-minded way.  Any improvements/suggestions are welcome in this review.

### DIFF
--- a/scripts/spheral-env.in
+++ b/scripts/spheral-env.in
@@ -2,5 +2,5 @@
 
 source @CMAKE_INSTALL_PREFIX@/.venv/bin/activate
 python "$@"
+exit $?
 deactivate
-
--- a/scripts/spheral-env.in
+++ b/scripts/spheral-env.in
@@ -2,5 +2,6 @@
 
 source @CMAKE_INSTALL_PREFIX@/.venv/bin/activate
 python "$@"
-exit $?
+exit_result=$?
 deactivate
+exit $exit_result


mdavis36:
I think we could write the return value to a variable and exit at the end with the result?

MishaZakharchanka (commented):
As far as I understand, activate and deactivate are there to make the interactive experience better. There are no lasting negative effects if we don't deactivate and the environment will be discarded when it finishes running anyway.

But I'm ok with either approach.

mdavis36:
@jmikeowen , @MishaZakharchanka Building and running this through the ats suite is returning 9 failed tests from: testPolygon, testPolyhedron, testNestedGridNeighbor, testTreeNeighbor, test_uniform_random, Colliding-Plates-1d, TensileRod-1d. Looking into this...

jmikeowen:
Yep, I'm seeing those failures as well.  It looks like these tests have been broken for a while, but we didn't realize it due to this exit code hiding problem.

mdavis36 (approved):

MishaZakharchanka (approved):

PR #82, opened 2021-08-04
Fixing testPolygon tests.

mdavis36:
# Summary
This fixes `closestPointOnFacet` and `closestPointAboveFacet` tests in `testPolygon`

- The `closestPointOnFacet` test was failing because `closestPointOnSegment` calculation had an issue which caculated the point in the opposite direction of the facets position in some cases giving `0.5*(p2+p1)` instead of `0.5*(p1+p2)` as it is calculated in `GeomFacet2D::position`
- The test for `closestPointAboveFacet` was broken because the point would be so far away from it that it would often find a point on the wrong facet.

### DIFF
--- a/src/Geometry/tests/testPolygon.py
+++ b/src/Geometry/tests/testPolygon.py
@@ -305,6 +305,7 @@ class TestPolygon(unittest.TestCase):
     # Closest point on facets.
     #---------------------------------------------------------------------------
     def testClosestPointOnFacets(self):
+
         facets = self.polygon.facets
         for f in facets:
             p = f.position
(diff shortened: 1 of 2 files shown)


jmikeowen:
This is a good catch, but the fix for pointDistances isn't quite right I think.  The functions "closestPointOnSegment" should return a point on the line segment between (a0, a1), but the original version and fixed here both have bugs.  On lines 30 and 47 we need to constrain the multiplier ptest to the range [0, a01mag] to ensure the result is on the line segment (a0,a1), so these lines should be changed to

a0 + std::max(0.0, std::min(a01mag, ptest))*ahat

jmikeowen (approved):
Looks good.

PR #83, opened 2021-08-05
test_uniform_random fix.

mdavis36:
Fixes for test_uniform_random. Adding a couple of missing default values necessary and small fixes in the class.


jmikeowen (approved):
Looks fine -- though I think these changes are already rolled into the Neighbor test fix pull request.

PR #84, opened 2021-08-05
Added missing timers for fsisph.

brbass:
The definitions for fsisph timers are missing, which causes a runtime error with missing symbols when building with debug.


jmikeowen (approved):
These look good.  Be sure and let Jason know about changes like this.

jmikeowen:
I'm gonna go ahead and merge this, since this is an important build fix.

PR #85, opened 2021-08-05
Fix for Neighbor tests

mdavis36:
Fixes the erroneous test for Neighbor tests. It appears this particular test was incomplete, I think this is might be what was in mind...


jmikeowen (approved):
This looks good -- my only request would be to remove the print statement on line 379 of NeighorTestBase.py.  That's the remnant of debug printing, and isn't necessary for normal test running.

PR #87, opened 2021-09-21
graphics flag caused ATS fail w/out X-forwarding terminal

jmpearl:
ATS was failing on the CollisionlessSphereCollapse test. Graphics output was throwing when x-forwarding wasn't enabled.


jmikeowen (approved):
These ATS tests should always disable graphics, so this was an oversight.

PR #89, opened 2021-09-28
Sod test generalization

jmpearl:
The baseline sod tests were modified to allow gamma to vary between the two gases. New 1D and 2D sod tests were added for stiffened gas eos (defaulting to a psuedo water gas interaction). The 1D stiffenedEOS sod test was added to the ats. Some updates for compatibility with the latest spheral  version are included.

SodWG_Example


jmikeowen (approved):
Looks fine

PR #92, opened 2021-09-29
Changes for Blast

brbass:
The largest change in this is to include the standard NodeLists in the connectivity calculation. There are several other changes here that support using a DataBase with different NodeLists than the Registrar, that add missing headers, and that make conduit build using static libraries similarly to axom.


cmake/tpl/qhull.cmake
```diff
@@ -3,7 +3,11 @@ set(QHULL_URL "https://github.com/qhull/qhull/archive/2019.1.tar.gz")
 set(QHULL_CACHE "${CACHE_DIR}/2019.1.tar.gz")
 set(QHULL_SRC_DIR ${QHULL_PREFIX}/src/qhull/src)
 
-set(${lib_name}_libs libqhullstatic.a)
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(${lib_name}_libs libqhullstatic_d.a)
+else()
+  set(${lib_name}_libs libqhullstatic.a)
+endif()
```
mdavis36, line 10: I can't find anything in the README for qhull that mentions libqhullstatic_d.a, I only see libqhullstatic_r.a, but I'm assuming it exists when you need to build this in debug. The TPL scripts currently build all of our TPLs as release by default. Do you find you need to build local debug TPLs like this often?

cmake/tpl/conduit.cmake
```diff
@@ -6,9 +6,9 @@ set(CONDUIT_CACHE "${CACHE_DIR}/${CONDUIT_DIST}")
 list(APPEND ${lib_name}_INCLUDES $<BUILD_INTERFACE:${${lib_name}_DIR}/include/${lib_name}>)
 
 set(${lib_name}_libs 
-    libconduit.so
-    libconduit_blueprint.so
-    libconduit_relay.so
+    libconduit.a
+    libconduit_blueprint.a
+    libconduit_relay.a
```
mdavis36, line 11: I like this fix over changing the library rpaths. We will need to rebuild the LC TPL directories first before merging this however.

src/Distributed/DistributedBoundary.cc
```diff
@@ -1034,10 +1034,10 @@ cullGhostNodes(const FieldList<Dimension, int>& flagSet,
...
   const int procID = domainID();
 
-  typedef NodeListRegistrar<Dimension> Registrar;
-  Registrar& registrar = Registrar::instance();
-  CONTRACT_VAR(registrar);
-  REQUIRE((int)numNodesRemoved.size() == registrar.numNodeLists());
+  // typedef NodeListRegistrar<Dimension> Registrar;
+  // Registrar& registrar = Registrar::instance();
+  // CONTRACT_VAR(registrar);
+  // REQUIRE((int)numNodesRemoved.size() == registrar.numNodeLists());
```
mdavis36, line 1040: Any reason to keep this around?

src/Boundary/Boundary.cc
```diff
@@ -95,15 +95,14 @@ void
 Boundary<Dimension>::cullGhostNodes(const FieldList<Dimension, int>& flagSet,
                                     FieldList<Dimension, int>& old2newIndexMap,
                                     vector<int>& numNodesRemoved) {
-
-  auto& registrar = NodeListRegistrar<Dimension>::instance();
-  REQUIRE((int)numNodesRemoved.size() == registrar.numNodeLists());
+  const auto& nodeListPtrs = flagSet.nodeListPtrs();
+  REQUIRE(numNodesRemoved.size() == nodeListPtrs.size());
+  // auto& registrar = NodeListRegistrar<Dimension>::instance();
+  // REQUIRE((int)numNodesRemoved.size() == registrar.numNodeLists());
```
mdavis36, line 101: I noticed when building in debug w/ the C++ libs as static that this would fail, the fix for registrar is addressed in an upcoming PR but I'm wondering if you saw this fail too?

mdavis36 (approved):
I like this, some of the fixes here are things I've seen recently too, we will need to wait on merging this until the LC TPL's are updated due to the change in conduit.

cmake/tpl/qhull.cmake
brbass: In Blast when we build debug, all the packages are built in debug, including qhull. When debugging, it makes things much easier, since you can dive into the third-party libraries. I can't remember where I say libqhullstatic_d, but if I run into it again I'll send you a link.

cmake/tpl/conduit.cmake
brbass: This is also how Blast installs the libraries and is consistent with what axom does in Spheral. I'm not sure why the defaults are different between the two packages.

src/Distributed/DistributedBoundary.cc
brbass: No, no reason to keep it around. I can remove it.

src/Boundary/Boundary.cc
brbass: It built fine for me in debug mode. Maybe it is a case where the library name changes or something as is the case for qhull?

MishaZakharchanka (approved):

jmikeowen (approved):
Looks good to me

PR #95, opened 2021-10-05
Axom mpi build

mdavis36:
# Summary
Fix to remove the patch step when installing Axom.
TPLs have bee rebuilt on LC for TOSS gcc 8.3.1 and builds Spheral succesfully with this fix.

### TODO

- [x] LC TOSS gcc 8.1.0 is in progress still needs to be rebuilt, having issues linking with the LC hdf5 install


jmikeowen (approved):
I was surprised at first the number of files changed, but I see it's all just replacing "mpi.h" -> <mpi.h>.  Looks good.

PR #99, opened 2021-10-14
Feature/pmin damage

jmikeowen:
This brings in a few changes:

- Most importantly, I've added a new member to SolidEquationOfState (minimumPressureDamage), which is the minimum pressure that damaged material should support.  This replaces the old binary flag in solid hydros (negativePressureInDamage), so that flag has been removed.  This new variable is the minimum pressure that should be allowed for fully damaged material, and defaults to 0.0 (i.e., no tension is allowed).  This allows us more control over the new feature in the code where damaged material is able to maintain strength properties, rather than behaving purely as a fluid.
- I found that the ANEOS third party library comes with three provided material inputs (quartz, dunnite, and serpentine), so I've added a convenient constructor for ANEOS that allows the user to specify one of those in concise syntax.  I also modified the aneos.cmake TPL build prescription to install these pre-packaged inputs, and have updated the TOSS gcc 8.1/8.3 pre-built LC third party installs accordingly.
- I improved some of our Python "shadow" code (factory functions that present a nicer Python front-end to the user), including stuff that now preserves the C++ pybind11 documentation in the help() for such functions.  We really need to package up these sorts of utilities into something designed to standardize writing shadow interfaces like this, since this is a nice method of quickly giving users a much friendlier interface to the code.

### DIFF
--- a/src/Utilities/tests/testBiQuadraticInterpolator.py
+++ b/src/Utilities/tests/testBiQuadraticInterpolator.py
@@ -15,6 +15,25 @@ from mpl_toolkits.mplot3d import Axes3D
 import random
 rangen = random.Random()
 
+#===========================================================================
+# A generator for creating a range of x values to test
+#===========================================================================
+def xygen(n, xmin, xmax, ymin, ymax):
+    assert n > 4
+    count = 0
+    while count < n:
+        if count == 0:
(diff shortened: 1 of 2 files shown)


jmikeowen:
Spoke with Paul and he's good to go with these changes.

PR #102, opened 2021-12-09
Bug fix/gsph units error

jmpearl:
For the edge case of zero sound speed the HLLC solver will break. To avoid this it switches to a normal av implementation when c->0. There's a division operation in there that should have been multiplication so the effective AV pressure would have had some really bizarre units. I think it didn't shown up because the av switch is pretty rare and was only activated on tests where the density was conviently equal to one.


jmikeowen (approved):
Oh yeah, gotta love those corner cases.  This looks fine, for code changes.  Do we have any tests that exercise this switch and would show the bad Q behavior?

PR #103, opened 2021-12-09
Fsisph diffuse mod

jmpearl:
Version of record for the FSI paper. 

1) second order diffusion terms 
2) gave eval derivs its own file to be more constistent w/ other hydros
3) converted slidesurface to an av like implementation to make it more user-friendly
4) updated some tests to have FSI


jmikeowen (approved):
This is looking good.

PR #104, opened 2021-12-11
Fix to just install files under version control for tests directory

jmikeowen:
This is a bugfix so that the install stage for tests only copies files under version control, rather than scraping everything from the tests source directory.

### DIFF
--- a/cmake/SetupSpheral.cmake
+++ b/cmake/SetupSpheral.cmake
@@ -160,8 +160,26 @@ endif()
 if (ENABLE_TESTS)
   add_subdirectory(${SPHERAL_ROOT_DIR}/tests/unit/CXXTests)
 
-  install(DIRECTORY ${SPHERAL_ROOT_DIR}/tests/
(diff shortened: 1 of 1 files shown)


mdavis36:
@jmikeowen This looks okay to me assuming it runs with the test suite fine from the install location after installing to an empty directory @MishaZakharchanka do you have time to give this a go?

MishaZakharchanka (commented):
Looks good, but we need to exclude the runCXXTests.ats file from the install or the CXX tests wont work in the ats. I'll look into how to do this Monday morning.

jmikeowen:
Added logic to avoid copying the tests source directory version of runCXXTests.ats, since that's a special case of the build system generating the version we want to use.

mdavis36 (approved):

MishaZakharchanka (approved):
Looks good, sorry I wasn't able to get to this. My power was out today from about 10am until 4pm.

PR #105, opened 2021-12-12
Update KelvinHelmholtz-2d_McNally.py

jmpearl:
mixing scale tracker had a typo in the reduced density -- rhoprof.  it as using "nodes" from the last iteration of the loop at ln 364 instead of nodeL from the local loop.


jmikeowen (approved):
Good catch -- what do you see in the output now with this fix?

jmpearl:
I didn't actually compare the output ke before/after I bet its a noticeable change though. I stumbled upon it when I was trying to do a mass-match distribution and the different nodelist resolutions were causing a run-time error

PR #93, opened 2021-10-01
Spack TPLs & GitLab-CI

mdavis36:
# Summary
This PR brings in gitlab-ci testing along with changes to the TPL management system for maintaining our TPL builds across various compilers and configurations.

- TPLs are to be managed through `spack` removing the need to maintain our own TPL build system for each of our dependencies.
- A set of dev-tool scripts are supplied to help Spheral developers with managing Spack for Spheral builds. These can be found at `scripts/dev-tools`
- Gitlab will run a suite of tests on LLNL CZ machines for PR/branches created by verified LLNL developers of the Spheral project. 
  - TOSS
    - gcc@8.3.1^mvpich2 
    - gcc@8.1.0^mvapich2 Debug
    - gcc@8.3.1~mpi
    - clang@9.0.0^mvapich2
  - BLUEOS
    - gcc@8.1.0^spectrum-mpi
    - gcc@8.1.0~mpi Debug  

## Other Changes
- Axom, Conduit and QHull now use shared libs.
- Fixes Clang Builds

## TODO
- [x] Documentation pass
- [x] Build mirrors in shared space for RZ and CZ users.
- [x] Tag release of Spheral before merging. That will be the l$t supported version of Spheral with the old TPL system. This PR will most likely break builds if a user tries to use the old TPL system.

PR #106, opened 2022-01-21
Fixing Debug & MPI-Off builds

mdavis36:
This PR is the same as #88 without the changes for gitlab-ci / spack.

We should merge this before merging #93 so we can tag a version of Spheral with these fixes included.


mdavis36:
Tested w/ debug, mpi=off, and release+mpi=on all tests pass.

jmikeowen (approved):
Looks good

PR #111, opened 2022-02-03
FSISPH update

jmpearl:
updating how we get scalar damage from tensor damage in FSISPH to be more consistent with SolidSPHHydro. 

I also rolled an update on the specificThermalEnergyPolicy into this PR. GSPH and FSISPH were using the same policy duplicated in both subdirectories. I removed the redundant files and moved the policy over into the hydro dir w/ the rest of the policies.


jmikeowen (approved):
Looking good -- I'm all for consolidating/removing redundant code.

PR #113, opened 2022-02-07
Dem

jmpearl:
Discrete element method package. This is an implementation of a linear-damped spring flavor of DEM from Schwartz 2012 & Zhang 2017. Friction is modeled with a linear-spring + damper with sliding friction element. A shape factor is used to model non-sphericity and cohesion. 

It includes the option to form composite particles by assign overlap to explicitly model non-sphericity. This feature is still experimental though so user be warned.


jmikeowen (dismissed):
This is very good work Jason.  I see several places you had to generalize our infrastructure since DEM is not a fluid or solid (like the Integrator and NodeGenerators).  I'm sure we'll wind up fiddling with how to handle those things as we figure out how to run hybrid cases.  Awesome work getting this capability in!

mdavis36:
Please add some notes in RELEASE_NOTES about the changes here, and wait for #151 to go to develop before merging.

mdavis36 (approved):

jmikeowen (approved):
Reapproving with latest changes

PR #118, opened 2022-03-10
Intel OneAPI Builds / CI Support 

mdavis36:
# Summary
Adding Intel OneAPI builds to gitlab CI

This PR includes:
  - New Compiler definitions for oneapi 2022.1 on Toss LC systems
  - Gitlab CI job for oneapi build.
  - Fix / patch for opensubdiv with newer intel / oneapi compilers.
  - Turning off bitwise comparisons on some tests, which pass floating point check but not bit for bit tests.
  - `Compiler.cmake` from old PR #59 to have somewhere to place compiler specific flags


jmikeowen:
My one concern is turning off the bit for bit checks on our tests.  Do we know why Intel is failing that?

jmikeowen:
Ah, I suspect I know what's wrong.  These are the tests comparing with reference data generated by a different compile configuration.  Perhaps we should just add reference data generated by this compiler choice and have this option check against the correct one.

mdavis36:
> Ah, I suspect I know what's wrong. These are the tests comparing with reference data generated by a different compile configuration. Perhaps we should just add reference data generated by this compiler choice and have this option check against the correct one.

Yeah I thought it was something along those lines I had a quick go at playing with some optimization flags to see if that did anything, but I couldn't get them to match. The OneAPI compiler is still relatively new to me so I'm not sure what exactly could be producing this difference. I think it would be good to have some sort of ats flag for enabling or disabling bitwise checks and only performing them on our most common compiler configurations, but that is out of the scope of this PR for now.

jmikeowen:
Is there a simple way to check the compiler we used from Python?  I know Python distuils provides that kind of information, so long as they compiler used for Python matches our general build compiler.  Then we could have compiler dependent reference data for these tests.

I'm uncomfortable just turning off the reference data checking for these tests because
- these tests don't have analytic solutions, and without the reference check we're really not checking anything
- these tests have caught and squashed unintended consequences and bugs in the past

mdavis36:
Understandable, we could implement a configurable definition / static header file. Many projects do this that are built with CMake to help encode compiler / build information from CMake into the project e.g. [RAJA](https://github.com/LLNL/RAJA/blob/develop/include/RAJA/config.hpp.in), [CAMP](https://github.com/LLNL/camp/blob/main/include/camp/config.in.hpp), [Umpire](https://github.com/LLNL/Umpire/blob/develop/src/umpire/config.hpp.in). We could then add a python interface for querying this information of course. Also we could pass our compiler `definitions` currently implemented [here](https://github.com/LLNL/spheral/blob/develop/cmake/CMakeDefinitions.cmake) into this header file.

jmikeowen:
We talked about this already, but for completeness here I should add a comment.  I'd like to add some ability to distinguish the compiler and make our reference data tests like those removed here TensileRod-1d and CollidingPlates use that to discriminate how to pass rather than disabling them.

mdavis36:
@jmikeowen @MishaZakharchanka This PR is ready to go, please have a look over when you get a chance.

mdavis36:
@jmikeowen @MishaZakharchanka This PR is ready to go, please have a look over when you get a chance.

MishaZakharchanka (approved):

jmikeowen (approved):
Looks good to go again.  

One suggestion we may want to consider -- perhaps we should add the compiler info to the block printed by "import Spheral" (that includes stuff like git hash and branch).  This might help us when we need to look at problems users send us with their run log.   I don't think we should hang up this pull request for such a change, but perhaps we might want to add that in a separate feature branch.

PR #120, opened 2022-03-10
Bugfix/redistribution notification

jmikeowen:
Adding pedantic checking for expired pointers to the RestartRegistrar and RedistributionNotificationRegistrar, which ensures we don't try to call into objects that have been deleted.  This fixes a memory error Jason was encountering.


jmpearl (approved):
that fixed it

mdavis36 (approved):

PR #121, opened 2022-03-11
GSPH revamp

jmpearl:
Reworked the GSPH module's architecture. GSPH now has its own pure-virtual GenericHydro class. this contains the bulk of the code for these Riemann-based hydro classes. The GSPHHydroBase and MFMHydroBase specialize off the generic with the main difference between the two coming from their different eval-derivs methods and a density/volume based approach to continuity respectively.

The changes were needed to get some more flexibility out of the derivative definition we use to reconstruct the interface state in the Riemann solvers. The changes also make the GPSH / MFM specializations a little cleaner I think.


jmikeowen (approved):
Very impressive getting these new options in and supported.

PR #123, opened 2022-03-31
updating CullenDehnen

jmpearl:
CullenDehnen artificial viscosity package was throwing a seg fault. Needed to have the DvDt hydrofieldname updated for modern spheral. All the other changes are aesthetic: made a bunch of types auto, added R and v_sig as fieldlists to remove the call to newFluidFieldList in the FinalizeDerivs method, and parsed out the set-get methods into an inline file.

### DIFF
--- a/src/ArtificialViscosity/CullenDehnenViscosity.cc
+++ b/src/ArtificialViscosity/CullenDehnenViscosity.cc
@@ -54,6 +54,8 @@ CullenDehnenViscosity(ArtificialViscosity<Dimension>& q,
   mCullAlpha2(FieldStorageType::CopyFields),
   mDalphaDt(FieldStorageType::CopyFields),
   mAlphaLocal(FieldStorageType::CopyFields),
+  mR(FieldStorageType::CopyFields),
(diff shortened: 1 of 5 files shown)


mdavis36:
I would vote to revert the auto changes if at all possible. It makes understanding the code far easier understand for non physics people/ future developers who will have to work with this. If some of the templated types here are too verbose, type aliases would be preferred at the very least.

jmpearl:
ahh roger that yeah I'll roll that part back

jmpearl:
Alright types are back in there, Mike, I ran it on the bolide problem and the Cl_multiplier field is passing the smell test. Take a look at the enrollDerivatives method though, I added resizes for mDalphaDt and mAlphaLocal not sure if that was an unnecessary addition?

jmikeowen:
Ah, good work catching this was broken.  I'm guessing we're not using this in any of our integration tests.  Maybe we should add a version of one of our standards (Noh or Sod) that uses this option to check it doesn't get our of date and broken again.

jmikeowen (approved):
I see our checks have passed, so good to go.

PR #124, opened 2022-04-06
bugfix/NohAnalyticSolution

jmpearl:
the analytic soln. needed a could edits for off-nominal initial conditions


jmpearl:
Tested it on against CRKSPH in 2d with different radial inflow velocities and everything looks good. 1D looked clean too.

jmikeowen (approved):
Always good to catch these inadvertently hardwired things.

PR #125, opened 2022-04-15
Bugfix/no mirrors

mdavis36:


mdavis36:
@jmpearl @jmikeowen I've turned off OneApi Builds on this PR as I haven't been able to find an acceptable workaround for the issues I'm seeing yet. This PR also fixes the issue Jason found earlier today. I accidentally deleted my fork of Uberenv (once my PR there merged), however the updated Uberenv with my changes is here. Assuming everything comes back green on this we should be able to merge this and fixes the two fires I started.

jmikeowen (approved):
Looks like the checks have come back OK, and the changes look good to me.  Bummer about the OneAPI compiler problems.  We'll have to revisit that in the future, but I agree with suspending those tests for now.

mdavis36:
@jmikeowen It is unfortunate. I managed to get our TPLs and Spheral to build, but then ran into an undefined symbols runtime error in the mpi fortran library. I couldn't find those symbols in any other mpi fortran implementation so I'm not sure what was going on there and ran out of time. Definitely need to ping some people about this on LC and see if we can get some help with it because I'm at a loss.

PR #129, opened 2022-04-29
FSISPH-update

jmpearl:
This one's a combination of the strength in damage update and an update to how slides are implemented

1) Slides -- all of the interface fields: interfaceNormals, interfaceSmoothness, and interfaceFraction; are now housed in the hydro class and treated as any other fieldlist updated w/ evaluate derivs. The SlideSurface class is now a bit lighter, containing only helper methods. This cuts the number of pair-loops and allows corrected kernels to be used for our interface fields.

2) Strength in damage -- FSI hadn't been fully updated for Spheral's new strength in damage approach. It was still assuming fully damaged material was zero-strength. This PR changed FSISPH to handle same-material damaged strength similarly to spheral's default solid SPH solver. The biggest difference in the FSI implementation is that the stress tensor is only damage-reduced if cauchy stess is tensile in the direction of the pair-wise line of action. For different materials we're assuming zero adhesion for the time being.

other minor changes: 

3) xsph in FSI was changed to only be applied in negative pressure regions to fight tensile instability

4) added a new fsi density summation method based on weighting by Hdeti/Hdetj at interfaces

5) enabled fragIDs to activate FSI's different material treatment for nodes belonging to the same nodeList


jmikeowen (approved):
Looking good

PR #130, opened 2022-04-29
Feature/sidre file io spio

MishaZakharchanka:
This PR will add SPIO to be used in SidreFileIO

- allows a variable amount of files to write the restart (0 < file count <= number of ranks)
- variable in controller constructor to specify how many files you want to write
-- default = number of ranks
-- variable does nothing if restartFileConstructor != SidreFileIO
- functional tests added to Noh-planar-1d.py
- Fixes an existing bug in SidreFileIO that would cause crashes when using more than 2 MPI ranks

There are also minor changes to other existing parts of SidreFileIO, but nothing functional


src/SimulationControl/SpheralController.py
```diff
@@ -28,6 +28,7 @@ def __init__(self, integrator, kernel,
                  restartBaseName = "restart",
                  restartObjects = [],
                  restartFileConstructor = SiloFileIO,
+                 restartFileCount = 0,
```
jmikeowen, line 31: What does restartFileCount == 0 mean in general?  Do we want this to default to mpi.procs?

MishaZakharchanka: I set it in SidreFileIO.hh. I figured that it should be set/changed by the class that uses it. Also could make the controller less confusing when using other FileIO, but could be changed if needed.

MishaZakharchanka: I set it in SidreFileIO.hh. I figured that it should be set/changed by the class that uses it. Also could make the controller less confusing when using other FileIO, but could be changed if needed.

jmikeowen: I think I have two suggestions here:
- the name "restartFileCount" implies (to me) the total number of restart files in a directory, not per timeslice.  Maybe change the parameter name to something like "numRestartFilesPerTimeslice" or something similarly horribly long and descriptive.  At least in the SpheralController.
- I'd go ahead and make the default mpi.procs as well, since I think that's less confusing.  We can then screen for setting it to something else, and have the SpheralController issue a warning if someone is setting it to something other than mpi.procs and not using SidreFileIO.

mdavis36: I think renaming this to something like `SPIOFileCount` might be more descriptive as this is only applicable to when we want to tune SPIO. Defaulting this to `None` might also make things clearer. We can then default to always using the `(filename, access)` constructor even with Sidre. Then using `(filename, access, count)` when `SPIOFileCount` is not `None` and we are using Sidre). This would help us avoid duplicating the default semantics  both in the C++ interface and the python one.

tests/functional/Hydro/Noh/Noh-planar-1d.py
```diff
@@ -177,6 +179,7 @@
             dataDirBase = "dumps-planar-Noh",
             restartBaseName = "Noh-planar-1d",
             restartFileConstructor = SiloFileIO,
+            restartFileCount = mpi.procs,
```
mdavis36, line 182: As per my comment above setting this to `None` stops us from having to define the default behavior again, unless we want to explicitly enforce some other value here.

tests/functional/Hydro/Noh/Noh-planar-1d.py
```diff
@@ -17,8 +17,10 @@
...
+#ATS:t12 = test(       SELF, "--graphics None --clearDirectories True  --checkError True  --dataDir 'dumps-planar-sidre-parrallel' --restartStep 20 --restartFileConstructor SidreFileIO", np=2, label="Planar Noh problem -- 1-D (parallel) with Sidre")
+#ATS:t13 = testif(t12, SELF, "--graphics None --clearDirectories False --checkError False --dataDir 'dumps-planar-sidre-parrallel' --restartStep 20 --restartFileConstructor SidreFileIO --restoreCycle 20 --steps 20 --checkRestart True", np=2, label="Planar Noh problem -- 1-D (parallel) RESTART CHECK with Sidre")
+#ATS:t14 = test(       SELF, "--graphics None --clearDirectories True  --checkError True  --dataDir 'dumps-planar-spio' --restartStep 20 --restartFileConstructor SidreFileIO --restartFileCount 1", np=6, label="Planar Noh problem -- 1-D (parallel) with Sidre (SPIO check)")
+#ATS:t15 = testif(t14, SELF, "--graphics None --clearDirectories False --checkError False --dataDir 'dumps-planar-spio' --restartStep 20 --restartFileConstructor SidreFileIO --restartFileCount 1 --restoreCycle 20 --steps 20 --checkRestart True", np=6, label="Planar Noh problem -- 1-D (parallel) RESTART CHECK with Sidre (SPIO check)")
```
mdavis36, line 23: Just to clarify, `t12` and `t13` here are using SPIO here correct? and we are forcing `t14` and `t15` to run on a single file source?

src/SimulationControl/SpheralController.py
```diff
@@ -552,7 +555,10 @@ def dropRestartFile(self):
         import time
         start = time.clock()
         fileName = self.restartBaseName + "_cycle%i" % self.totalSteps
-        file = self.restartFileConstructor(fileName, Create)
+        if self.restartFileConstructor is SidreFileIO:
+            file = self.restartFileConstructor(fileName, Create, self.restartFileCount)
+        else:
+            file = self.restartFileConstructor(fileName, Create)
```
mdavis36, line 561: As above, default to `(filename, access)` here even with Sidre. Use `(filename, access, count)` only when `restartFileCount` is not `None` and we are using `Sidre`

src/FileIO/SidreFileIO.cc
```diff
@@ -136,6 +136,30 @@ SidreFileIO::SidreFileIO(const std::string fileName, AccessType access):
...
+//------------------------------------------------------------------------------
+// Construct with a set number of restart files.
+//------------------------------------------------------------------------------
+SidreFileIO::SidreFileIO(int numFiles):
+  FileIO(),
+  mDataStorePtr(0)
+{
+  if (numFiles > 0 && numFiles <= Process::getTotalNumberOfProcesses())
+    numRestartFiles = numFiles;
+}
```
mdavis36, line 148: I might be missing it, is there anywhere we are using this constructor, or a instance we might use it? Or is this an overload of a base FileIO constructor?

mdavis36 (changes requested):

tests/functional/Hydro/Noh/Noh-planar-1d.py
jmikeowen: I'm good with these two changes (SPIOFileCount and None as default).  None is consistent with how we're defaulting other similar parameters here.

src/SimulationControl/SpheralController.py
mdavis36: I was doing my review as you left your last comment @jmikeowen so I didnt' see it, changing the name to either or a combo of the two if fine with me :)

tests/functional/Hydro/Noh/Noh-planar-1d.py
MishaZakharchanka: Yes, we use Spio all the time if we built with mpi. Otherwise it's normal Sidre.

src/FileIO/SidreFileIO.cc
MishaZakharchanka: It's not being used anywhere, but I added both constructors if needed for future use somewhere. I can remove and it can be added later if needed.

mdavis36: Okay lets go ahead and remove it provided it doesn't break anything to do so.

src/SimulationControl/SpheralController.py
```diff
@@ -55,6 +56,8 @@ def __init__(self, integrator, kernel,
         self.kernel = kernel
         self.restartObjects = restartObjects
         self.restartFileConstructor = restartFileConstructor
+        if restartFileConstructor is SidreFileIO: # FileCount only applies to SidreFileIO
```
mdavis36, line 59: I don't think we need this condition anymore

src/SimulationControl/SpheralController.py
```diff
@@ -584,6 +595,8 @@ def loadRestartFile(self, restoreCycle,
         if self.restartFileConstructor is GzipFileIO:
             file = self.restartFileConstructor(fileName, Read)
                                                #readToMemory = True)
+        elif self.restartFileConstructor is SidreFileIO and self.SPIOFileCountPerTimeslice is not None:
+            file = self.restartFileConstructor(fileName, Read, self.SPIOFileCountPerTimeslice)
         else:
             file = self.restartFileConstructor(fileName, Read)
```
mdavis36, line 601: I think this is left over from something else but it looks like GzipFileIO falls under the else category here now.

mdavis36 (changes requested):
Just two more quick things

src/SimulationControl/SpheralController.py
MishaZakharchanka: Is that something we want to get rid of now or in its own PR for tracking?

mdavis36: I think it's fine to do here unless @jmikeowen has any objections

jmikeowen: Are you asking if we want to get rid of GzipFileIO?  My preference is to keep it -- it's a functional FileIO implemenation that only depends on Python (used to be our default in fact).  You never know when we might be on a platform where Silo or Sidre aren't available (or those could go away).  I'd need to know why we want to remove it before I'd go for it.

mdavis36: No, not remove GzipFileIO. We are just calling the same constructor signature in `if GzipFileIO` as we are in `else` so we could just have that condition fall under else.
```
if self.restartFileConstructor is SidreFileIO and self.SPIOFileCountPerTimeslice is not None:
  file = self.restartFileConstructor(fileName, Read, self.SPIOFileCountPerTimeslice)
else:
  file = self.restartFileConstructor(fileName, Read)
```

mdavis36 (approved):
Looks good, thanks @MishaZakharchanka !

jmikeowen (approved):
Agreed, looking good to go.

PR #133, opened 2022-05-09
Feature/spack cuda support

mdavis36:
# Summary
This PR adds functionality to the Spheral build system to compile with NVCC. 
  - This PR also brings in changes from #148 
  - Documentation updates can be found here : 
       https://spheral.readthedocs.io/en/feature-spack-cuda-support/index.html

### Additions
  - Adds a CUDA smoke test that can be called from Spheral python API.
  - Adds gitlab-ci jobs on blueos systems to build with NVCC / CUDA 11.
 
### Changes
  - Spheral requires C++ 14.
  - ATS filters / options are injected into ./spheral-atstest script (more below).
  - C++ flag suppression only happens when building C++ objects.
  - Always point to locally installed ATS executable, instead of default lcats version.

### TPL Changes
  - Bumps ATS version to 7.0.9 This requires a variety of other changes.
    - A Final ATS report script is required after calling ATS in our Gitlab-CI (LCATS / ATS doesn't seem to report failed tests through exit codes anymore). This inspects the ATS log files for FAILED Spheral tests and returns an exit code to our CI system.
    - Generates the necessary ATS filters and extra options required for a specific Spheral build and injects the options into spheral-atstest script (Issues on blueos systems with quotation marks and < symbols).
    - numpy 1.16.6 -> 1.16.5 because ATS 7.0.9 requires less than 1.16.6...
  - Eigen is bumped to 3.4.0 for a fix when building with cuda code. The previous version was accessing legacy CUDA headers.
   - HDF5 needs to be built with/without MPI support depending on Axom MPI required support.


mdavis36:
@MishaZakharchanka @jmikeowen This is ready for a review when you get a chance 😄

MishaZakharchanka (approved):

src/ArtificialViscosity/ArtificialViscosity.hh
```diff
@@ -180,12 +180,12 @@ public:
   virtual void restoreState(const FileIO& file, const std::string& pathName);
   //****************************************************************************
 
-protected:
-  //--------------------------- Protected Interface ---------------------------//
   // Allow descendents to request that sigma be calculated.
   bool calculateSigma() const;
   void calculateSigma(bool value);
 
+protected:
```
jmikeowen, line 187: Why was it necessary to move the sigma functions to public from protected?

mdavis36: It was previously, let me check with the latest library changes since we are no longer building the python libs with nvcc.

mdavis36: Reverted this change.

jmikeowen (approved):
Yep, looking good to go to me.

MishaZakharchanka (approved):

PR #137, opened 2022-05-18
Hotfix for gitlab / permission issues in LLNLSpheral

mdavis36:
This PR templates out our gitlab-ci configs so that we can hopefully use them across both projects and fixes some path issues with the new generated config file.


jmikeowen (approved):
Looking good to me.

PR #139, opened 2022-06-03
Bugfix/bulk moduli

jmpearl:
This PR updates a couple of the EOSs to enforce more consistent behavior between the bulk sound speed and bulk modulus.


jmikeowen:
The only thing holding this request back for me is the change to the PorousEOS bulk modulus.  The existing method scales the bulk modulus consistently with how the pressure is treated.  The proposed change also explicitly overrides whatever the base solid EOS wants to set the bulk modulus to to be rho*cs^2.  What is wrong with the original method?  And do we always want to ignore what the underlying solid EOS is doing in it's bulk modulus implementation?

jmikeowen (approved):
Good to go.  I'd suggest we get together (maybe myself, Jason, and Katie) to talk more about the porous EOS behavior.  I see what you guys are concerned about with the bulk modulus and sound speed.  I'd like to know how this is handled in other codes like iSALE perhaps, which has the same porous model.

jmpearl:
Same here, on iSALE / other codes strain porosity.  I could throw something on the calendar for next week

PR #143, opened 2022-06-10
Feature/spherical sph

jmikeowen:
This pull request brings in the spherical SPH formalism, which is currently working with both fluid and solid standard SPH.  We have added ATS tests of the new formalism in both these regimes.


src/Neighbor/NestedGridNeighbor.hh
```diff
@@ -243,6 +243,13 @@ private:
   static const int mGridNormalMagnitude;
 };
 
+//------------------------------------------------------------------------------
+// Initialize static variables.
+//------------------------------------------------------------------------------
+template<typename Dimension> const double NestedGridNeighbor<Dimension>::ln2inverse = 1.0/log(2.0);
+template<typename Dimension> const int NestedGridNeighbor<Dimension>::mEndOfLinkList = -1;
+template<typename Dimension> const int NestedGridNeighbor<Dimension>::mGridNormalMagnitude = 1024;
+
```
mdavis36, line 252: Same here.

src/Physics/GenericHydro.cc
```diff
@@ -162,6 +164,14 @@ dt(const DataBase<Dimension>& dataBase,
   // Initialize the return value to some impossibly high value.
   auto minDt = make_pair(std::numeric_limits<double>::max(), string());
 
+  // Define a function for computing the velocity divergence (different for curvilinear coordinates)
+  std::function<double(const Tensor&, const Vector&, const Vector&)> Fdiv = [](const Tensor& DvDxi, const Vector& posi, const Vector& veli) { return DvDxi.Trace(); };
```
mdavis36, line 168: I'm usually all for type verbosity, but we can make this `auto` if you'd like.

src/SPH/SPHHydroBase.hh
```diff
@@ -217,6 +217,17 @@ public:
...
+  template<typename KernelType>
+  void evaluateDerivativesImpl(const Scalar time,
+                               const Scalar dt,
+                               const DataBase<Dimension>& dataBase,
+                               const State<Dimension>& state,
+                               StateDerivatives<Dimension>& derivatives,
+                               const KernelType& W,
+                               const KernelType& WQ,
+                               const TableKernel<Dimension>& WT) const;
+
```
mdavis36, line 230: So pretty much just moved SPHEvaluateDerivatives into SPHHydroBase, correct? Or are there some other more detailed changes that are harder to see with the git diffs?

src/Utilities/BiCubicInterpolatorInline.hh
```diff
@@ -0,0 +1,431 @@
...
+    1.0, x2, x2*x2, x2*x2*x2, y3, x2*y3, x2*x2*y3, x2*x2*x2*y3, y3*y3, x2*y3*y3, x2*x2*y3*y3, x2*x2*x2*y3*y3, y3*y3*y3, x2*y3*y3*y3, x2*x2*y3*y3*y3, x2*x2*x2*y3*y3*y3,
+    1.0, x3, x3*x3, x3*x3*x3, y3, x3*y3, x3*x3*y3, x3*x3*x3*y3, y3*y3, x3*y3*y3, x3*x3*y3*y3, x3*x3*x3*y3*y3, y3*y3*y3, x3*y3*y3*y3, x3*x3*y3*y3*y3, x3*x3*x3*y3*y3*y3,
+    1.0, x4, x4*x4, x4*x4*x4, y3, x4*y3, x4*x4*y3, x4*x4*x4*y3, y3*y3, x4*y3*y3, x4*x4*y3*y3, x4*x4*x4*y3*y3, y3*y3*y3, x4*y3*y3*y3, x4*x4*y3*y3*y3, x4*x4*x4*y3*y3*y3,
+
+    1.0, x1, x1*x1, x1*x1*x1, y4, x1*y4, x1*x1*y4, x1*x1*x1*y4, y4*y4, x1*y4*y4, x1*x1*y4*y4, x1*x1*x1*y4*y4, y4*y4*y4, x1*y4*y4*y4, x1*x1*y4*y4*y4, x1*x1*x1*y4*y4*y4,
+    1.0, x2, x2*x2, x2*x2*x2, y4, x2*y4, x2*x2*y4, x2*x2*x2*y4, y4*y4, x2*y4*y4, x2*x2*y4*y4, x2*x2*x2*y4*y4, y4*y4*y4, x2*y4*y4*y4, x2*x2*y4*y4*y4, x2*x2*x2*y4*y4*y4,
+    1.0, x3, x3*x3, x3*x3*x3, y4, x3*y4, x3*x3*y4, x3*x3*x3*y4, y4*y4, x3*y4*y4, x3*x3*y4*y4, x3*x3*x3*y4*y4, y4*y4*y4, x3*y4*y4*y4, x3*x3*y4*y4*y4, x3*x3*x3*y4*y4*y4,
+    1.0, x4, x4*x4, x4*x4*x4, y4, x4*y4, x4*x4*y4, x4*x4*x4*y4, y4*y4, x4*y4*y4, x4*x4*y4*y4, x4*x4*x4*y4*y4, y4*y4*y4, x4*y4*y4*y4, x4*x4*y4*y4*y4, x4*x4*x4*y4*y4*y4;
```
mdavis36, line 72: How many tries did this take? 😱

mdavis36 (commented):
Absolutely incredible amount of work here. well done @jmikeowen! I have a few comments and questions on my first pass of this.

src/Gravity/TreeGravity.hh
```diff
@@ -240,6 +240,15 @@ private:
...
 
+//------------------------------------------------------------------------------
+// Define our static members.
+//------------------------------------------------------------------------------
+template<typename Dimension> unsigned TreeGravity<Dimension>::num1dbits = 21U;
+template<typename Dimension> uint64_t TreeGravity<Dimension>::max1dKey = 1U << TreeGravity<Dimension>::num1dbits;
+template<typename Dimension> uint64_t TreeGravity<Dimension>::xkeymask = (1U << TreeGravity<Dimension>::num1dbits) - 1U;
+template<typename Dimension> uint64_t TreeGravity<Dimension>::ykeymask = TreeGravity<Dimension>::xkeymask << TreeGravity<Dimension>::num1dbits;
+template<typename Dimension> uint64_t TreeGravity<Dimension>::zkeymask = TreeGravity<Dimension>::ykeymask << TreeGravity<Dimension>::num1dbits;
+
```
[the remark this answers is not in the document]
jmikeowen: Yeah, that's what I thought, but this was required by Clang 9.0.  I looked up the rules, and it turns out we need them in the header because if you have static members whose values depend on the template parameter they have to be declared in the header so every compilation unit knows the correct value for the template parameter being used.  I'd be happy to move them back to the .cc files if we can, but I according to what I read Clang is correct and we have to keep 'em in the headers.

src/SPH/SPHHydroBase.hh
jmikeowen: Yeah, this is mostly just collapsing the evaluateDerivatives back into the .cc file since we have so many SPH variants now.  I split this out years ago to help with parallel building, but the new Cmake system parallelizes the build quite well so it's no longer necessary.

There are a couple of other minor changes in the implementation due to changes in the kernel interface I implemented getting the SphericalKernel done.  Nothing substantive.

src/Utilities/BiCubicInterpolatorInline.hh
jmikeowen: No comment!  :)

src/Physics/GenericHydro.cc
jmikeowen: I'd like to, but the compiler can't seem to figure it out if I do.

jmikeowen: This looks like our last issue -- I've tried auto (first thing I tried when I first wrote this in fact), but it doesn't compile.   I think the problem is that I want to conditionally assign to this variable in the succeeding if tests, but the type it resolves to with auto isn't allowing that.  This awful std::function declaration gives me an assignable type I can use.  Any suggestions to improve it?

jmikeowen: Following suggestions from @mdavis36 , this can now be expressed as an auto statement, so the latest commit resolves this request.

mdavis36 (approved):

PR #148, opened 2022-07-08
Single C++ lib

mdavis36:
# Summary
This PR is a refactor of the C++ Spheral library structure.

## Design Review
This PR refactors our current C++ libraries to be defined in CMake as OBJECT libraries (A collection of compiled object files and headers without a linked library). These object libraries are then linked into a single `SpheralCXX` library.

This is required to solve a number of runtime anomalies found when running the python interface of Spheral. These anomalies commonly occurred when building Spheral C++ libs as static.

### Other Wins
 - The single C++ library also seems to solve some of out TPL shared/ static library issues:
   - Can use static Conduit and Qhull (will be resolved in future PRs...)
 - Less C++ libs to link to for customer apps.
 - Smaller overall C++ lib size as everything is linked down into one file.

### DIFF
--- a/cmake/Compilers.cmake
+++ b/cmake/Compilers.cmake
@@ -48,7 +48,10 @@ if (NOT ENABLE_MISSING_INCLUDE_DIR_WARNINGS)
 endif()
 message("-- Compiler missing include dir warnings ${ENABLE_MISSING_INCLUDE_DIR_WARNINGS}")
 
+set(CUDA_WARNING_FLAGS -Xcudafe=\"--diag_suppress=esa_on_defaulted_function_ignored\")
+
 add_compile_options("$<$<COMPILE_LANGUAGE:CXX>:${CXX_WARNING_FLAGS}>")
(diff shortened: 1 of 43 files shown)


jmikeowen:
This looks like a nice change for some of our customers!  Good work sorting this out for NVCC.

One question: I see that we still have spheral_add_cxx_library and you made a new cmake function for the OBJ libraries we're building now.  Do we still use spheral_add_cxx_library anywhere?

mdavis36:
> This looks like a nice change for some of our customers! Good work sorting this out for NVCC.
> 
> One question: I see that we still have spheral_add_cxx_library and you made a new cmake function for the OBJ libraries we're building now. Do we still use spheral_add_cxx_library anywhere?

Yep we're using it to build the single Spheral_CXX library.

jmikeowen:
> Yep we're using it to build the single Spheral_CXX library.

Ah, makes sense!

MishaZakharchanka (approved):

brbass (approved):

PR #149, opened 2022-07-12
Docker & Azure-CI

mdavis36:
# Summary
- This PR adds a Docker recipe to build Spheral on a fresh install of Ubuntu 20.04.
- Adds an azure test so we can automate testing Spheral for Ubuntu 20.04.
  - Azure builds are configured to run only when changes are pushed to develop, this should give us reasonable tracking for when a new feature / PR breaks external builds without requiring an Azure build for every PR.
- spheral-atstest renamed spheral-lcatstest.
- spheral-atstest is now for external users / running ats without lcats.


jmikeowen (approved):
Looks good to me.

PR #151, opened 2022-07-18
pbatch gitlab-ci jobs

mdavis36:
Hotfix to get our Ruby gitlab-ci jobs running in pbatch instead of pdebug.


jmikeowen (approved):
I'm surprised we don't have to execute some batch command to get the pbatch allocation, but if this works it's great!  Once the tests pass it looks ready to go to me.

PR #152, opened 2022-07-26
ale3d

ptsuji:
Branch to sync with ALE3D.


src/Utilities/SpheralCTimers.cc
```diff
@@ -0,0 +1,15 @@
...
+Timer TIME_SpheralC                                ("Root Timer for SpheralC");
+Timer TIME_SpheralC_initialize                     ("SpheralPsuedoScript::initialize", TIME_SpheralC);
+Timer TIME_SpheralC_updateState                    ("SpheralPsuedoScript::updateState", TIME_SpheralC);
+Timer TIME_SpheralC_initializeBoundariesAndPhysics ("SpheralPsuedoScript::initializeBoundariesAndPhysics", TIME_SpheralC);
+Timer TIME_SpheralC_initializeStep                 ("SpheralPsuedoScript::initializeStep", TIME_SpheralC);
+Timer TIME_SpheralC_evaluateDerivatives            ("SpheralPsuedoScript::evaluateDerivatives", TIME_SpheralC);
+Timer TIME_SpheralC_iterateHfield                  ("SpheralPsuedoScript::iterateHfield", TIME_SpheralC);
+Timer TIME_SpheralC_computeFragmentID              ("SpheralPsuedoScript::computeFragmentID", TIME_SpheralC);
+Timer TIME_SpheralC_sampleLatticeMesh              ("SpheralPsuedoScript::sampleLatticeMesh", TIME_SpheralC);
+Timer TIME_SpheralC_polyhedralMesh                 ("SpheralPsuedoScript::polyhedralMesh", TIME_SpheralC);
```
mdavis36, line 15: We are using Caliper now for our Timer implementation. I don't think these are supported any longer

ptsuji:
Anything else I need to do on this pull request? The only test that's failing is testSidreFileIO on one of the ATS runs (blueos_gcc_8_3_1_~mpi_Debug_test):

https://lc.llnl.gov/gitlab/spheral/spheral/-/jobs/936753

mdavis36:
> Anything else I need to do on this pull request? The only test that's failing is testSidreFileIO on one of the ATS runs (blueos_gcc_8_3_1_~mpi_Debug_test):
> 
> https://lc.llnl.gov/gitlab/spheral/spheral/-/jobs/936753

Nope, I'm re-running that build and test, it _should_ pass.

mdavis36 (approved):

mdavis36:
@jmikeowen @ptsuji Once https://rzlc.llnl.gov/gitlab/owen/LLNLSpheral/-/pipelines/119145 passes we can merge this.

src/Damage/ProbabilisticDamageModel.cc
```diff
@@ -140,7 +140,7 @@ initializeProblemStartup(DataBase<Dimension>& dataBase) {
       mInitialVolume(i) = mass(i)/rho(i) * mVolumeMultiplier;
       mVmin = std::min(mVmin, mInitialVolume(i));
       mVmax = std::max(mVmax, mInitialVolume(i));
-      Key seedi = mSeed;
```
jmikeowen, line 143: Why this change?  We use the Key typedef in case in the future (or on different architectures) we need to change this, so we can do it one place.  This change hardwires this choice only at this point.

src/CRKSPH/SolidCRKSPHHydroBaseRZ.cc
```diff
@@ -161,11 +161,7 @@ initializeProblemStartup(DataBase<Dim<2> >& dataBase) {
     const unsigned n = mass[nodeListi]->numElements();
     for (unsigned i = 0; i != n; ++i) {
       const Scalar circi = 2.0*M_PI*abs(pos(nodeListi, i).y());
-#ifdef WIN32
```
jmikeowen, line 164: This (and the similar changes below for other RZ specializations) worry me.  We shouldn't be dividing by zero densities anywhere.  When is this happening? And why this protection for only the WIN32 platform?  If we really need to protect from div by zero we should use one of the safeInv functions (src/Utilities/safeInv.hh), but this shouldn't be happening at all.

src/Kernel/TableKernelAleInst.cc.py
```diff
@@ -13,12 +16,14 @@
   template class TableKernel< Dim< %(ndim)s > >;
 """
 
-for Wname in ("GaussianKernel",
```
jmikeowen, line 16: I think this instantiation file is only for use ALE3D, correct?  Perhaps we should move this out of the Spheral source entirely into a separate thing like SpheralC, since this isn't even built by Spheral's build system so far as I can tell.

src/Mesh/Mesh.cc
```diff
@@ -531,7 +531,9 @@ cleanEdges(const double edgeTol) {
         nodeMap[n2] = n1;
       }
     }
-    replace_if(nodeMask.begin(), nodeMask.end(), std::bind2nd(std::equal_to<unsigned>(), 2), 1);
```
jmikeowen, line 534: This sort of change has been made in a few places.  The lambda function is fine, but I'm curious why make the change?

jmikeowen (changes requested):
I added a few questions in individual file changes.  Hopefully this shows up since I'd like to clarify those before approving.

src/Mesh/Mesh.cc
ptsuji: This was a change to move to C++17, I believe.

src/Kernel/TableKernelAleInst.cc.py
ptsuji: Yes, this script is only used by ALE3D. We can put it in SpheralC if that's more appropriate.

src/Damage/ProbabilisticDamageModel.cc
ptsuji: It seems that the Key typedef does not play well with boost::hash_combine on MacOS, looking at the commit log. I just tried reverting it on my Mac laptop and got the following error:


In file included from /Users/tsuji1/Code/ale3d/imports/spheral/build-osx_gcc-relwithdebinfo/src/Damage/weibullFlawDistributionOwenInst2d.cc:6:
/Users/tsuji1/Code/ale3d/imports/spheral/src/./Damage/weibullFlawDistributionOwen.cc: In instantiation of 'Spheral::Field<Dimension, std::vector<double> > Spheral::weibullFlawDistributionOwen(unsigned int, double, double, const FluidNodeList<Dimension>&, int, double, const Field<Dimension, int>&) [with Dimension = Dim<2>]':
/Users/tsuji1/Code/ale3d/imports/spheral/build-osx_gcc-relwithdebinfo/src/Damage/weibullFlawDistributionOwenInst2d.cc:15:108   required from here
/Users/tsuji1/Code/ale3d/imports/spheral/src/./Damage/weibullFlawDistributionOwen.cc:87:26: error: no matching function for call to 'hash_combine(Key&, long long unsigned int&)'
   87 |       boost::hash_combine(seedi, keyList(0,i));
      |       ~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
In file included from /Users/tsuji1/Code/ale3d/lib/osx_gcc/relwithdebinfo/BOOST/1.74.0b/compiler-gcc12.2.0/cpp-14/include/boost/functional/hash.hpp:6,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Mesh/MeshConstructionUtilities.hh:23,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Mesh/MeshInline.hh:15,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Mesh/Mesh.hh:252,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./NodeList/SmoothingScaleBase.hh:13,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./NodeList/FluidNodeListInline.hh:3,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./NodeList/FluidNodeList.hh:122,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Field/NodeIteratorBaseInline.hh:4,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Field/NodeIteratorBase.hh:84,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Field/NodeIterators.hh:4,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Field/FieldInline.hh:4,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Field/Field.hh:256,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./NodeList/NodeListInline.hh:2,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./NodeList/NodeList.hh:229,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Utilities/globalNodeIDsInline.hh:14,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Utilities/globalNodeIDs.hh:63,
                 from /Users/tsuji1/Code/ale3d/imports/spheral/src/./Damage/weibullFlawDistributionOwen.cc:6:
/Users/tsuji1/Code/ale3d/lib/osx_gcc/relwithdebinfo/BOOST/1.74.0b/compiler-gcc12.2.0/cpp-14/include/boost/container_hash/hash.hpp:413:17: note: candidate: 'template<class T> void boost::hash_combine(std::size_t&, const T&)'
  413 |     inline void hash_combine(std::size_t& seed, T const& v)
      |                 ^~~~~~~~~~~~
/Users/tsuji1/Code/ale3d/lib/osx_gcc/relwithdebinfo/BOOST/1.74.0b/compiler-gcc12.2.0/cpp-14/include/boost/container_hash/hash.hpp:413:17: note:   template argument deduction/substitution failed:
/Users/tsuji1/Code/ale3d/imports/spheral/src/./Damage/weibullFlawDistributionOwen.cc:87:27: note:   cannot convert 'seedi' (type 'Key' {aka 'long long unsigned int'}) to type 'std::size_t&' {aka 'long unsigned int&'}
   87 |       boost::hash_combine(seedi, keyList(0,i));
      |                           ^~~~~

src/CRKSPH/SolidCRKSPHHydroBaseRZ.cc
jmikeowen: Ah yes, this was just for the mysterious extra elements we were getting on Fields with the Windows compiler. Since this is removing that extra hack, good to go.

src/Kernel/TableKernelAleInst.cc.py
jmikeowen: Yep, we should probably move this.  Not urgent right now, but in the near future.

src/Mesh/Mesh.cc
jmikeowen: Ah, OK.  The lambda is actually more readable, so this should be fine.

src/Damage/ProbabilisticDamageModel.cc
jmikeowen: Ah, in this case I'd suggest we make this change with an #Ifdef __APPLE__ to restrict to only that platform.  My worry is if in the future we change KeyTraits::Key to something incompatible with size_t in some way, this could turn into a subtle and hard to find bug.  We could either change just this line in this way, or change type overall type of KeyTraits::Key in src/Utilities/KeyTraits.hh.

ptsuji: I've added the ifdefs in both ProbabilisticDamageModel.cc and weibullFlawDistributionOwen.cc to only use std::size_t on Mac.

ptsuji:
@jmikeowen I think I've resolved all the issues you mentioned, can you change the status of your review?

jmikeowen (approved):
Yep, this change resolves my concerns, so we should be ready to go.

PR #155, opened 2022-08-05
Feature/caliper

MishaZakharchanka:
This PR adds Caliper to replace our existing Timer class.

- Timing zones that were previously defined with Timer class now use Caliper
- Adds a macro layer around the Caliper calls and uses the same ENABLE_TIMER flag that the Timer class used
- Adds documentation to provide usability similar to that of the Timer class as well as links to other useful queries that can be made from Caliper


scripts/spack/configs/toss_3_x86_64_ib/packages.yaml
```diff
@@ -130,6 +130,11 @@ packages:
     - spec: tar@1.26
       prefix: /usr
     buildable: false
+  elfutils:
+    externals:
+    - spec: elfutils@0.176
+      prefix: /usr
+    buildable: false
```
mdavis36, line 137: We'll need to add this to the bluesos variant to get this to build over there.

cmake/tpl/zlib.cmake
```diff
@@ -1 +1 @@
-set(${lib_name}_libs libz.a)
+set(${lib_name}_libs libz.so)
```
mdavis36, line 1: Why was this changed?

scripts/spack/packages/spheral/package.py
```diff
@@ -70,6 +70,7 @@ class Spheral(CachedCMakePackage, CudaPackage, PythonPackage):
     depends_on('py-sphinx@1.8.5', type='build', when='+docs')
     depends_on('py-sphinx-rtd-theme@0.5.0', type='build', when='+docs')
 
+    depends_on('caliper', type='build')
```
mdavis36, line 73: We should be able to use the static version of this library, try adding `~shared` to this spec. Then we can remove caliper from our rpaths and import `libcaliper.a`.

src/Utilities/timerLayer.hh
```diff
@@ -0,0 +1,16 @@
...
+#define TIME_BEGIN(regionName) CALI_MARK_BEGIN(regionName)
+#define TIME_END(regionName) CALI_MARK_END(regionName)
+
+#else // TIMER
+// Stub TIME macros, when TIME is zero
+#define TIME_FUNCTION
+#define TIME_BEGIN(regionName)
+#define TIME_END(regionName)
+
+#endif // TIMER
```
mdavis36, line 16: Lets go ahead and have this moved to Utilities/Timer.hh I think it makes sense to keep all of our timer stuff in one place.

mdavis36 (changes requested):
@MishaZakharchanka good work so far, just a few notes about tpls and build system stuff.

@jmikeowen Do we want to keep backwards compatibility with the old timer class. If there's reason to we can keep both and have a switch on the macro layer (we'll use Caliper as the default), or do we want to rip it out now?

cmake/tpl/zlib.cmake
MishaZakharchanka: There was a problem that I was running into while building a week ago and this was a simple solution to keep working. I need to check it out again.

src/Utilities/timerLayer.hh
MishaZakharchanka: Since the old timer was completely removed, we could move/rename the new macro-layer to Timer.hh, I'm just not sure if there is a naming convention we could follow for macro classes.

mdavis36: Often for MACROS it can be useful to prefix them with the project name for possibly overloaded name, I think in this case the macro names are fine for us.

Lets move this to Timer.hh since this is our only interface now.

docs/developer/dev/diagnostic_tools.rst
```diff
@@ -0,0 +1,41 @@
...
+So far there are two different types of regions in Spheral, using the following macros:
+::
+
+  TIME_FUNCTION
+  TIME_BEGIN("timer_name")
+  TIME_END("timer_name")
+
+- ``TIME_FUNCTION`` can be added to the very beginning of a function and creates a region for the entire function using the function's name. ``TIME_FUNCTION`` uses just the function name and no class or parameter information, so be careful when using this method with functions that could share names.
+
+- ``TIME_BEGIN("timer_name")`` and ``TIME_END("timer_name")`` create a region between the two different calls and use the string (in this case timer_name) as the name. This is the most similar method to the removed Spheral::Timer.
```
mdavis36, line 41:
```suggestion
- ``TIME_BEGIN("timer_name")`` and ``TIME_END("timer_name")`` create a region between the two different calls and use the string (in this case timer_name) as the name.
```

mdavis36 (changes requested):
Couple of comments, we should also add `-DENABLE_TIMER=On` to one of our CI tests. Lets use `cuda_11_gcc_8_3_1_spectrum` in `spheral/.gitlab/specs.yml`

MishaZakharchanka:
One more thing, that could be changed that I ran into after making the macro layer. I often found myself writing TIMER_BEGIN() and not TIME_BEGIN(), not sure if it's more intuitive to have the former.

jmikeowen (approved):
This all looks pretty good to me.  One question: I see we just use strings passed to the start/stop calls to define which region is being timed.  This seems potentially fragile, as if someone mistypes the string in the start and stop calls, they will be mismatched.  Is there someway we find and identify incorrect patterns like this?  It might be possible by examining the output of the timers or something?

MishaZakharchanka:
Any input errors I tried while I was testing (for example forgetting to close a section) are caught by Caliper during compile time. I don't remember if I gave this specific scenario a try, so I'll give it a shot.

MishaZakharchanka:
If there is a mismatch in the names for the beginning and end of a marked region caliper will output errors that look like this:
`== CALIPER: (0): region nesting error: trying to end "CheapRK2" but current region is "CheapRK2222"`

This happens when you try to query Caliper, not during compilation like I previously stated.

mdavis36 (approved):

PR #156, opened 2022-08-08
Bugfix/dem

jmpearl:
Fix for rotational momentum convservation issue


jmikeowen (approved):
Looks good to me -- glad to see you added the test to catch this in the future.

PR #157, opened 2022-08-09
Bugfix/probabilistic damage model

jmikeowen:
This is mostly a minor Python binding bugfix for ProbabalisticDamageModel.  

I have also updated some of the Python scripts for plotting our flaw distributions created by these models as well, but that's more of a documentation/example thing.


MishaZakharchanka (approved):

jmikeowen:
Gonna pull the trigger on this one, since it's a bug fix and it's pretty minor.

PR #159, opened 2022-09-01
Feature/polyhedral gravity models

jmpearl:
Added a approximate polyhedral gravity model that can be used to model the gravitational fields of asteroids/comets. The model using numerical integration performed over a 3d surface mesh representing the body's surface. The current version uses a really simple centroidal quadrature rule to perform the integration which isn't the most accurate but should be good for our purposes.  I added an ats test which also surves as an implementaiton example. 

This PR also contains some small modifications to some of the Geometry classes.


jmikeowen (approved):
This is looking great!  I see one test failed in the CI, and it looks like one of our spurious tolerance problems so I've resubmitted that test.  Once we verify it passes we can merge.

PR #160, opened 2022-09-09
Bugfix/llnl spheral lib fix

jmikeowen:
This pull request makes some changes to our new method of building the C++ source into a single library.   That still works as before by default, but these changes allow us to as needed make more libraries that can be linked together in the end.  This is necessary for the LLNL extensions libraries to work, since the original method would create too many object files to be linked into a single library simultaneously.


MishaZakharchanka (approved):

mdavis36:
@jmikeowen a thought before we get ready to merge this. Do we still want to make multiple C++ libs now that we have resolved the duplicating TPLs issue? I don't think having multiple libs for the C++ component of Spheral is necessary anymore.

jmikeowen:
Currently we still only build one C++ library (libSpheral_CXX.so/a) as in develop, but we have the capability to build more than one.  In LLNLSpheral we are currently using this to build an additional C++ library (libSpheral_LLNLCXX.so/a), which links to the Spheral one.

I'm inclined to keep that model for now, but if we wind up with device code calls in LLNLSpheral we may have to do something different.  I agree that most likely we can go back to building a single library with all our compiled source in it (include LLNL extensions) with your fixes.  What do you think, would you like to switch to that approach?

mdavis36 (approved):
Lets go ahead and get this merged in, if we need to go back to LLNLSpheral only compiling one library then we can tackle that when it comes.

PR #162, opened 2022-09-15
Update MaterialPropertiesLib.py

jmpearl:
Basalt's reference density was a bit low -- Katy


jmikeowen (approved):
This looks like a good catch.

PR #165, opened 2022-09-26
Refactor/caliper optional

MishaZakharchanka:
This will make Caliper an optional tpl that can be added during the spack tpl setup. Once Caliper is added using tpl manager, ENABLE_TESTS will be turned on.

### DIFF
--- a/cmake/tpl/caliper.cmake
+++ b/cmake/tpl/caliper.cmake
@@ -1,1 +1,3 @@
-set(${lib_name}_libs libcaliper.a)
+if(ENABLE_CALPIER)
+  set(${lib_name}_libs libcaliper.a)
+endif()
--- a/cmake/tpl/caliper.cmake
+++ b/cmake/tpl/caliper.cmake
@@ -1,3 +1,3 @@
-if(ENABLE_CALPIER)
+if(ENABLE_TIMER)
   set(${lib_name}_libs libcaliper.a)
 endif()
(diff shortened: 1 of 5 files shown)


scripts/spack/packages/spheral/package.py
```diff
@@ -29,6 +29,7 @@ class Spheral(CachedCMakePackage, CudaPackage, PythonPackage):
     variant('openmp', default=True, description='Enable OpenMP Support.')
     variant('docs', default=False, description='Enable building Docs.')
     variant('shared', default=True, description='Build C++ libs as shared.')
+    variant('caliper', default=False, description='Build Caliper support.')
```
mdavis36, line 32: I don't think we need to make caliper/timers a variant in our spack package just yet. I think we can continue to always build it as a dependency. Locking in our host-configs to enabling / disabling timers seems explicit and I'd rather leave it as a CMake configuration option only so developer don't need to separate tpl-manager instances when building with and without timer support.

cmake/tpl/caliper.cmake
```diff
@@ -1 +1,3 @@
-set(${lib_name}_libs libcaliper.a)
+if(ENABLE_TIMER)
+  set(${lib_name}_libs libcaliper.a)
+endif()
```
mdavis36, line 3: We should move this logic to cmake/InstallTPLs.cmake so as to not even look for a caliper package

scripts/spack/packages/spheral/package.py
MishaZakharchanka: So we want the our tpls to always have Caliper, but we want the ENABLE_TIMER option to control not only enabling the Caliper code inside of our timer macro layer but also if Spheral looks for Caliper as a tpl? If I am understanding your comment correctly.

mdavis36: Yep, that is correct, if we aren't using timers we don't need to link or include caliper

jmikeowen (approved):
Looks like it should fix the problem!

mdavis36:
Go ahead and update LLNLSpheral with develop so we can test with the latest CI changes once everything comes back green there I will approve

mdavis36 (approved):

PR #168, opened 2022-10-11
fixing DEM restart issue

jmpearl:
This is a fix for a restart issue with DEM  in which the timestep -> infty. The issue would occur if the script was set up to skip node generation during restarts. DEM was simply setting the timestep on construction and without nodes the timestep calculation returns infty. To fix the issue, the timestep var is now a restartable var.

Updated 10/31/2022
In addition to the above, this PR also fixes conservation issues that would occur if the kullFrequency (now contactRemovalFrequency) was set to a number greater than 1. It also fixes conservation issues that would occur during redistribution. DEM should now be safe for arbitrary contactRemovalFrequency and with redistribution.

------
### ToDo :

### DIFF
--- a/src/DEM/DEM.py
+++ b/src/DEM/DEM.py
@@ -15,13 +15,20 @@ def LinearSpringDEM(dataBase,
                     staticFrictionCoefficient,
                     rollingFrictionCoefficient,
                     torsionalFrictionCoefficient,
-                    cohesiveTensileStrength=0.0,
-                    shapeFactor=0.0,
+                    cohesiveTensileStrength = 0.0,
+                    shapeFactor = 0.0,
                     stepsPerCollision = 25,
                     xmin = (-1e100, -1e100, -1e100),
                     xmax = ( 1e100,  1e100,  1e100)):
+
     assert dataBase.numDEMNodeLists == dataBase.numNodeLists, "all nodelists must be dem nodelists"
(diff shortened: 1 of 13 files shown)


jmikeowen:
Interesting -- so DEM timesteps are always constant?

We should add an ATS test that tests restart.  It looks like it should be simple enough to add those to the tests in tests/functional/DEM, kind of like we do with the dependent ATS restart tests in the planar Noh problem.  I can add those in this  branch if it's helpful.

mdavis36:
Will get you added to the sphapp user account so you can restart CI without a commit. Can you also add the ToDo section from [.github/pull_request_template.md](https://github.com/LLNL/spheral/blob/develop/.github/pull_request_template.md) to the PR description so we can track if the PR changes work with LLNLSpheral once this is ready.

jmikeowen:
We could add a restart test to impactingSquares-2d.py, which skips the problem generation when restarting.  I saw in develop that test will fail to restart properly.

jmikeowen:
Alternatively we could just modify one of the existing restart tests to skip the problem generation.

jmpearl:
sounds good, I'll throw in a restart test for the impacting squares

jmpearl:
Update: some of the issues mike found revealed some deeper issues in DEM w/ momentum conservation. Things are moving along but the PR still needs a little work.

jmikeowen (approved):
Good work tracking all this down Jason

PR #186, opened 2022-12-09
Feature/polyhedral gravity

jmikeowen:
# Summary

- This PR adds features and includes bugfixes
- It does the following:
  - New features/ API changes:
    * New polyhedral gravity solver
    * Improved DEM timestep choice, so that points that do not overlap do not overly constrain the timestep
    * Point potential gravity solver can now have non-unit metrics
    * Pair-max damage algorithm now decouples points in different NodeLists or with different fragmentIDs
    * Added direct support for FacetedVolumes to FileIO interface
    * ANEOS now allows shallow copies to be made
    * Polyhedron contain method can now optionally use Axom methods to test for containment
    * Adding user specified functions for shear modulus and yield strength as a function of damage    

  - Build changes / improvements:
    * Added ENABLE_NAN_EXCEPTIONS (default OFF) Cmake flag to raise an exception when a NAN occurs (Gnu only)

  * Bug Fixes / improvements:
    * Fixes for restarting without regenerating the original node positions in the Python script
    * Protected from division by zero in DEM when points coincide
    * Corrected support for minimum pressure (intact and damaged) with porous materials
    * Removed term driving damaged material to the reference density in solid hydros

------
### ToDo :


jmpearl (approved):
This is looking good. I had a couple questions more just out of curiousity than anything else: 

For the DEM timestep it looks like it always ends up with the dtSpring, were you putting in the infrastructure for later developement?

In ApproximatePolyhedralGravityModel was there an issue with the potential near the quadrature points? The 1/10th the quadrature point resolution (new ln. 84 in the .cc) is a bit larger than I usually run for the potential over in my grav code, but I've never run anything with so many particles near the surface -- maybe I should be running higher.

I didn't take a thorough look at the IO edits since I haven't spent much time in that part of the code.

jmikeowen:
@jmpearl , the time step in DEM is affected by the changes: the linear spring limit is only applied if the points overlap.  You can see this affects the results by running the impacting DEM test cases for instance, which now take fewer steps to run since the points start out non-overlapping, and separate after colliding.  I'd like to further change the timestep constraint down the line, but this is a start.

I don't recall why the change to the safety fudge factor in ApproximatePolyhedralGravityModel though!  We could probably put the old value back and be fine.

jmpearl (approved):

PR #192, opened 2023-01-10
Removing CXX_ONLY build guards

### LINKED ISSUE #181: Building C++ only turns off other features
At cmake/SetupSpheral.cmake::60, there are certain features of the code that are turned off when we build CXXONLY. These should be looked at to determine if this is necessary or its something that can be left to the users to decide while building.

MishaZakharchanka:
# Summary

- This PR is a refactor
- It does the following:
  - Modifies SetupSpheral.cmake removing the extra guards put in place when building CXX_ONLY
  - Fixes #181 

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#33)


jmikeowen (approved):
Ah, allowing the CXX tests to be built when building CXXONLY.  Looks good.

src/CXXTests/CMakeLists.txt
```diff
@@ -2,8 +2,6 @@ get_property(SPHERAL_CXX_LIBS GLOBAL PROPERTY SPHERAL_CXX_LIBS)
...
 add_subdirectory(DeviceTestLib)
 
-if(ENABLE_TESTS)
-  blt_add_executable( NAME spheral_cuda_test
-                      SOURCES Spheral_CUDA_Test.cc
-                      DEPENDS_ON Spheral_CXX ${spheral_blt_cxx_depends})
-endif()
+blt_add_executable( NAME spheral_cuda_test
+                    SOURCES Spheral_CUDA_Test.cc
+                    DEPENDS_ON Spheral_CXX ${spheral_blt_cxx_depends})
```
mdavis36, line 7: Do we want to remove the guard to build this target, this will always build now regardless of ENABLE_TESTS

MishaZakharchanka: I removed the guard because I was the one that added it, we could add it back if we see fit. I added it when I added CXX_ONLY build to disable building tests because of this test. I don't remember what the issue was, but there was some error that came up when building CXX_ONLY with this test before.

mdavis36: Let's keep it, if codes have an issue with this target for some reason it's easier to allow them to just turn it off.

jmikeowen (approved):
Looks good to me

mdavis36 (approved):

PR #201, opened 2023-03-18
hdf5 lib issue

mdavis36:
# Summary

- This PR is a bugfix
- It does the following:
  - Links Spheral to the shared version of HDF5 when building the python interface. This is to ensure consistent linking across all Spheral TPLs and h5py python module. (static libs can be forced for C++ only builds with ENABLE_STATIC_TPL).
  - Adds a github action to run workflows when changes are made in uberenv or spack related files of Spheral.

------
### ToDo :

### DIFF
--- a/.github/workflows/test-tpls.yml
+++ b/.github/workflows/test-tpls.yml
@@ -4,8 +4,8 @@ on:
   pull_request:
     paths:
       - .uberenv_config.json
-      - src/scripts/spack/**
-      - src/devtools/**
+      - scripts/spack/**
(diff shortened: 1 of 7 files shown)


jmikeowen:
It looks like ENABLE_STATIC_TPL is something we've only had defined in the conduit TPL setup to date.  Is this a switch we should pay attention to more broadly?  Seems like we should document it in the build_guide rst documentation.

jmikeowen (approved):
Looking good to go to me

PR #205, opened 2023-03-30
Flat connectivity updates and axom 0.7.0

brbass:
This adds one function and lots of comments to the FlatConnectivity class. 

This also includes the necessary changes for axom 0.7.0, which is what we are using in Blast. ~~ALE3D uses axom 0.5.0, so we may need to figure out how to handle a version difference there.~~

Edit: ALE3D is also on Axom 0.7.0. https://github.com/LLNL/spheral/pull/205#issuecomment-1490805187


ptsuji:
ALE3D is already on Axom 0.7.0 (commit 3d77e46). I can try building with these changes.

src/KernelIntegrator/FlatConnectivity.cc
```diff
@@ -20,6 +20,22 @@
 #include "Utilities/DBC.hh"
 #include "Utilities/globalNodeIDs.hh"
 
+namespace { // anonymous
```
jmikeowen, line 23: This method basically repeats what you can get from ConnectivityMap::numNeighborsForNode, and to some extent the DataBase method DataBase::numNeighbors (which computes a FieldList with the result for all points).  Wouldn't the method already in ConnectivityMap be good enough here though?

jmikeowen:
Overall looking fine to me.  I have one question about FlatConnectivity above, and we should get up to date with develop and confirm the CI is passing, but then I think this is ready to go.

cmake/InstallTPLs.cmake
```diff
@@ -58,8 +58,8 @@ Spheral_Handle_TPL(silo spheral_depends cxx)
 #Spheral_Handle_TPL(conduit spheral_depends cxx)
 find_package(axom REQUIRED QUIET NO_DEFAULT_PATH PATHS ${axom_DIR}/lib/cmake)
 if(axom_FOUND)
-  list(APPEND spheral_blt_cxx_depends axom fmt)
-  blt_patch_target(NAME fmt TREAT_INCLUDES_AS_SYSTEM On)
+  list(APPEND spheral_blt_cxx_depends axom axom::fmt)
+  blt_patch_target(NAME axom::fmt TREAT_INCLUDES_AS_SYSTEM On)
```
mdavis36, line 62: This fails as Axom at 0.7.0 does not export `fmt` as `axom::fmt`, only `fmt` and the `blt_patch_target` fails to find `axom::fmt`.

CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
 # CMakeLists to build the Spheral library.
 
 cmake_minimum_required(VERSION 3.10)
-project(spheral LANGUAGES CXX Fortran)
+project(spheral LANGUAGES C CXX Fortran)
```
mdavis36, line 4: I don't know if this belongs here, Spheral doesn't build any C targets as part of it's code base.

mdavis36 (changes requested):

CMakeLists.txt
brbass: I think this may have come from building with Blast, where Spheral failed to build without the additional C language. I want to say axom or umpire threw the error there. If we don't need this for the standalone Spheral build, though, we can make a patch to change this line in Blast and remove the C here.

mdavis36: Umpire would be the culprit of the two, they supply a C API. However we build our Umpire with the C interface enabled as well but it doesn't fail to build.

mdavis36 (changes requested):

cmake/InstallTPLs.cmake
brbass: Resolved by your commit below.

src/KernelIntegrator/FlatConnectivity.cc
```diff
@@ -787,6 +806,28 @@ globalOverlapNeighborIndices(const int locali,
...
 
+//------------------------------------------------------------------------------
+// Check whether NodeList ordering is appropriate for this function
+//------------------------------------------------------------------------------
+template<typename Dimension>
+bool
+FlatConnectivity<Dimension>::
+fluidNodeListsFirst(const DataBase<Dimension>& dataBase) const
+{
+  const auto numFluidNodeLists = dataBase.numFluidNodeLists();
```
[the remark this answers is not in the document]
brbass: Line removed.

mdavis36 (approved):
Looks good to me 👍

PR #243, opened 2023-10-19
Bugfix/demtests

jmpearl:
# Summary

- Sean caught some errors in the DEM test scripts not included in the ats. This PR updates those scripts.
- It does the following:
  - updates 3 DEM test scripts which were broken
  - Removes one duplicate DEM test script that was defunct
  - fixes the distance method of the RectangularPlanarSolidBoundary

------
### ToDo :


src/DEM/SolidBoundary/RectangularPlaneSolidBoundary.cc
```diff
@@ -37,9 +37,9 @@ distance(const Vector& position) const {
   const double signX = (q.x() > 0.0 ? 1. : -1.);
```
jmikeowen, line 37: Minor quibble, but these recreate the sgn() functions we provide in Utilities/SpheralFunctions.hh (I know, a very generic/non-descriptive file name).

jmikeowen:
Overall these changes look good.  My only request would be to replace the locally defined sign functions with the generic ones we provide in the Utilities directory.

src/DEM/SolidBoundary/RectangularPlaneSolidBoundary.cc
jmpearl: ah gotcha yeah no point in reinventing the wheel, I'll swap those out.

jmpearl: On second look, I realized I was just implementing bounds on that Vector's values in a confusing way

jmikeowen: Even better, looks good to me.

jmikeowen (approved):
Good to go

PR #246, opened 2023-11-21
P-alpha porosity

jmikeowen:
# Summary

- This PR is adds a new porosity model based on a P-alpha (pressure-distention) relationship.
- It does the following:
  - P-alpha and strain-alpha are now both supported.
  - We've added a new Porosity package with all porosity modeling support moved into this location.   Also a new PorosityModel base class for common features to P-alpha and strain-alpha.
  - Refactored how porosity is maintained in the State and participates in updating other physics variables.  The PorousEquationOfState and PorousStrengthModel are gone, though for backwards compatibility Python classes take their place and issue deprecation warnings.
  - Refactored the use of UpdatePolicies and how they relate to FieldLists.

------
### ToDo :


src/Material/GammaLawGas.cc
jmpearl, line 1: dPdrho needs an  ( i )

src/Material/IsothermalEquationOfState.cc
jmpearl, line 1: This is a bit irrelevant to the edits but notices that an external pressure is added to the bulk modulus but not the pressure, I think it should be the other way around?

src/Material/PolytropicEquationOfState.cc
jmpearl, line 1: Same comment as the IsothermalEOS

src/NodeList/SolidNodeList.cc
jmpearl, line 1: The longitudinal sound speed could be simplified to :
c_l = sqrt( (K + 4/3 * mu) /rho )

src/SolidMaterial/GruneisenEquationOfState.cc
jmpearl, line 1: should ln. 259 be like the following (switch to the 3rd power and throwing a 2 out front of the second term)?
(D*dAdrho - 2 * A*dDdrho)*FastMath::pow3(Dinv) + mb*eps);

jmpearl:
The DEM, FSISPH, and GSPH edits all look good

src/Strength/BulkModulusPolicy.cc
jmpearl, line 1: Looks like the FSIFieldNames include might not be needed?

src/Material/GammaLawGas.cc
jmikeowen: Good catch, fixed this

src/Material/IsothermalEquationOfState.cc
jmikeowen: Good catch -- applies to isothermal EOS as well.

src/Material/PolytropicEquationOfState.cc
jmikeowen: Yep, same thing.

src/NodeList/SolidNodeList.cc
jmikeowen: Are you sure about this one?  Wolfram-alpha doesn't find that solution.

jmikeowen: Yep, double checking it looks the most I can simplify to is

c_l^2 = (3*K + mu)*(3*K + 4*mu)/(27*K*mu)

I can go ahead and implement that unless I'm missing something?

jmikeowen: Oh never mind, I see you expanded with the multiplication by Youngs modulus.  I agree about the result of simplifying this.

src/Strength/BulkModulusPolicy.cc
jmikeowen: Agreed, removed that include

ldowen (dismissed):

scripts/spack/configs/x86_64/packages.yaml
```diff
@@ -10,11 +10,11 @@ packages:
   #    prefix: /usr/
   #  buildable: false
 
-  cmake:
```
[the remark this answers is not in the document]
mdavis36: This is for generic non LC systems. An example of which is the github-actions docker test we run where we build and test Spheral on Ubuntu 20. This file is based off of what is available in Ubuntu 20, in this PR the minimum CMake requirement was increased to 3.18 and so we are having spack build and install a newer version of CMake to use.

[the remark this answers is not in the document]
jmikeowen: So what do we want to do with this file?  Looks like this version does work with our Docker image (I'm assuming since the CI test passed).   Should we take these changes or revert to develop for this file?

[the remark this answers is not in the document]
mdavis36: If this works for our tests and your WSL2 instance then we should keep this in the meantime (no point holding up your dev workflow if it doesn't change anything). I'll open an issue about what we talked about today in regards to the external spack config files so that we can possibly revert them later for use with only our docker container.

.gitlab/machines.yml
```diff
@@ -10,6 +10,7 @@
     PARTITION: pdebug
     BUILD_ALLOC: srun -N 1 -c 36 -p pdebug -t 60
     TEST_ALLOC: ''
+    LC_MODULES: ""
```
mdavis36, line 13: The changes in this file can be removed as they were a redundant fix to the CUDA c++17 issue.

.gitlab/scripts.yml
```diff
@@ -13,7 +13,7 @@
     - echo $SPEC
 
     - ml load python/3
-    - $BUILD_ALLOC ./$SCRIPT_DIR/gitlab/build_and_install.py --spec="$SPEC" --extra-cmake-args="$EXTRA_CMAKE_ARGS"
+    - $BUILD_ALLOC ./$SCRIPT_DIR/gitlab/build_and_install.py --spec="$SPEC" --extra-cmake-args="$EXTRA_CMAKE_ARGS" --lc-modules="$LC_MODULES"
```
mdavis36, line 16: Same here, this can be reverted back.

mdavis36 (changes requested):
Just a couple of changes that need to be reverted, but after that I think this is good to go.

.gitlab/machines.yml
jmikeowen: Done in latest commit.

.gitlab/scripts.yml
jmikeowen: Same, done.

mdavis36 (approved):

jmpearl (approved):
Looks good to go!

ldowen (approved):

PR #247, opened 2023-11-27
Remove unnecessary forward declarations

adayton1:
# Summary

- This PR is a task
- It does the following:
  - Removes forward declarations when the actual declaration is already available
  - The old code was adding extra compiler overhead without any benefit

**Old pattern:**
```
#ifndef <class>_HH
#define <class>_HH

class <class> {
public:
  //--------------------------- Public Interface ---------------------------//

private:
  //--------------------------- Private Interface ---------------------------//
};

#else

// Forward declaration.
class <class>;

#endif
```

**New pattern:**
```
#ifndef <class>_HH
#define <class>_HH

class <class> {
public:
  //--------------------------- Public Interface ---------------------------//

private:
  //--------------------------- Private Interface ---------------------------//
};

#endif
```


jmikeowen:
Long ago this pattern was recommended to me (back in the late 90s), but if this is not needed then I'm all for removing this sort of forward declaration.

adayton1:
> Long ago this pattern was recommended to me (back in the late 90s), but if this is not needed then I'm all for removing this sort of forward declaration.

Yes, it's not adding any benefit (and may actually do some harm), so I think it should be removed.

mdavis36:
@adayton1 We're hoping to get this into the release this month. We've had some pretty big PRs merge into develop. Do you think you'd be able to get around to updating this branch within the next week or so? You would need to also propagate the changes here across the new files that have been added since those PRs.

adayton1:
> @adayton1 We're hoping to get this into the release this month. We've had some pretty big PRs merge into develop. Do you think you'd be able to get around to updating this branch within the next week or so? You would need to also propagate the changes here across the new files that have been added since those PRs.

Sorry, I've been out on vacation. I will take a look and get this up to date with develop.

adayton1:
@mdavis36, this should be up to date now.

adayton1:
I don't think the failing test is due to my changes. Does anyone know if RKInterpolation(RK interpolation - 2D septic) is expected to fail?

mdavis36:
> I don't think the failing test is due to my changes. Does anyone know if RKInterpolation(RK interpolation - 2D septic) is expected to fail?

We have some tests that will randomly fail due to calculating just outside tolerances. The RK Interpolation tests do this sometimes. I've restarted the test and it should pass

adayton1:
> > I don't think the failing test is due to my changes. Does anyone know if RKInterpolation(RK interpolation - 2D septic) is expected to fail?
> 
> We have some tests that will randomly fail due to calculating just outside tolerances. The RK Interpolation tests do this sometimes. I've restarted the test and it should pass

It passed this time. Thanks! Just need someone to review now.

mdavis36:
We need to make a PR on LLNLSpheral and test these changes there before merging, but I think this looks good to go once we get the green light over there.

jmikeowen (approved):
These changes look good to me.  Thanks for pitching in and helping clean stuff like this up!

PR #248, opened 2023-11-28
ale3d

ptsuji:
# Summary

- This PR is a mergeback of ALE3D changes
  - Syncs with what's on the current ALE3D develop branch
  - Remove Windows Visual Studio files
  - Add a field to mask certain SPH points from the computeFragmentField calculation
  - Fix in SolidSPHHydroBase to turn off tensile forces when fragments are separating (allows "bouncing")
  - Fix for time step estimate due to velocity divergence in RZ space
  - Change some boost::unordered_maps to std::unordered_map or std::map to build with nvcc


mdavis36 (changes requested):

src/NodeGenerators/compactFacetedVolumes.cc
```diff
@@ -40,7 +40,9 @@ unsigned compactFacetedVolumes(std::vector<typename Dimension::FacetedVolume>& s
 
   // Only proceed if there's work to do!
   int flagmax = *max_element(flags.begin(), flags.end());
```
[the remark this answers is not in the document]
ptsuji: Ok, I've moved this into the USE_MPI block.

build/spheral_ArtificialViscosity/spheral_ArtificialViscosity.vcxproj
adayton1, line 1: Does anyone besides us use these Windows project files? If not, I don't think they belong in the Spheral repository since I doubt they are general enough for anyone else to use.

src/Mesh/MeshInline.hh
```diff
@@ -369,7 +371,11 @@ minimumScale() const {
       result = std::min(result, (face.position() - zonePosition).magnitude2());
     }
   }
+#ifdef USE_MPI
```
adayton1, line 374: I wonder if there is a better pattern that hides all these ifdefs. Perhaps it would be simple enough to write functions like allReduceMax and allReduceMin that do this ifdef pattern behind the scenes.

```
double allReduceMin(double val) {
#ifdef USE_MPI
    return allReduce(val, MPI_MIN, Communicator::communicator());
#else
    return val;
#endif
}
```

src/Physics/GenericHydro.cc
```diff
@@ -168,7 +168,7 @@ dt(const DataBase<Dimension>& dataBase,
   if (GeometryRegistrar::coords() == CoordinateType::Spherical) {
     Fdiv = +[](const Tensor& DvDxi, const Vector& posi, const Vector& veli) { return DvDxi[0] + 2.0*veli[0]*safeInv(posi[0]); };
   } else if (GeometryRegistrar::coords() == CoordinateType::RZ) {
-    Fdiv = +[](const Tensor& DvDxi, const Vector& posi, const Vector& veli) { return DvDxi.Trace() + veli[0]*safeInv(posi[0]); };
```
adayton1, line 171: What was the reason for this change?

cmake/SetupSpheral.cmake
```diff
@@ -63,6 +63,10 @@ option(ENABLE_DEV_BUILD "Build separate internal C++ libraries for faster code d
 option(ENABLE_STATIC_CXXONLY "build only static libs" OFF)
 option(ENABLE_SHARED "Building C++ libs shared" ON)
 
+option(SPHERAL_ENABLE_ARTIFICIAL_CONDUCTION "Build artificial conduction package" ON)
```
adayton1, line 66: These are already added at line 58.

build/spheral_ArtificialViscosity/spheral_ArtificialViscosity.vcxproj
ptsuji: I don't think anyone else does, but it makes merging with the "pristine" branch less of a hassle.

cmake/SetupSpheral.cmake
ptsuji: Ok, I'll remove this then.

src/Physics/GenericHydro.cc
ptsuji: I reverted this to what it was before, removing the veli[0]*safeInv(posi[0]) term. The extra term crashes the Spheral time step constraint in our 2D RZ problems, where it was stable previously.

src/Mesh/MeshInline.hh
ptsuji: I think this kind of big change could be left for a future merge request.

build/spheral_ArtificialViscosity/spheral_ArtificialViscosity.vcxproj
ptsuji: I've removed the Windows project files.

adayton1 (approved):
LGTM

src/Utilities/Tree.hh
```diff
@@ -88,7 +88,7 @@ public:
 
   //--------------------------- Private Interface ----------------------------//
   // Define the types we use to build the tree.
-  typedef boost::unordered_map<CellKey, Cell> TreeLevel;
+  typedef std::unordered_map<CellKey, Cell> TreeLevel;
```
ptsuji, line 91: @mdavis36 @jmikeowen just a note, there are a few changes like this which Peter Robinson pushed last week to get Spheral to build with nvcc. Apparently it did not like the v1.74.0 implementation of boost::unordered_map.

mdavis36: @ptsuji We have been building Spheral with boost v1.74.0 and NVCC 11.1.1 for quite some time now. I'm not opposed personally to moving to the `std` implementation but I would be curious to know why that is failing for you and not us. What Cuda version are you building with?

ptsuji: We're building with cuda 11.8.0 (compiler is /usr/tce/packages/cuda/cuda-11.8.0/bin/nvcc). @robinson96 can you comment on what the issue with nvcc was?

robinson96: My best recollection is that the std lib for the version of gcc we were on did not have hash functions defined for some of the key types spheral was using.

src/Mesh/MeshInline.hh
mdavis36: The `allReduce` interface is already guarded by `USE_MPI`, the changes guarding all of these calls shouldn't be necessary. I think the issue here is that our usual thin interface for `communicator` when building without MPI support is also guarded out. Communicator headers should be included with or without MPI support in order to support the current `allReduce` interface.

mdavis36 (changes requested):

cmake/tpl/polytope.cmake
```diff
@@ -1 +1,10 @@
 set(${lib_name}_libs libpolytope.a)
+
+if(APPLE)
+  set(SHARED_EXT "dylib")
+  set(${lib_name}_libs libpolytope.${SHARED_EXT})
+endif()
+
+if(ENABLE_STATIC_TPL)
+  string(REPLACE ".${SHARED_EXT}" ".a;" ${lib_name}_libs ${${lib_name}_libs})
+endif()
```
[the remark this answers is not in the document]
ptsuji: Which macro do you want changed, the APPLE or SHARED_EXT?

ptsuji:
@mdavis36 for some reason I can't respond to your comment about allReduce; I've moved the Communicator.hh stuff to Utilities, and have removed all of the unnecessary ifdefs with USE_MPI.

src/DataBase/DataBase.cc
```diff
@@ -1923,7 +1938,6 @@ globalSamplingBoundingVolume(typename Dimension::Vector& centroid,
 				    xminNodes, xmaxNodes,
 				    xminSample, xmaxSample);
 
-#ifdef USE_MPI
```
ldowen, line 1926: @ptsuji @mdavis36 Should this ifdef be removed? It seems like it was here before and might be intentional.

mdavis36: This one is intentional and should be left in.

cmake/tpl/polytope.cmake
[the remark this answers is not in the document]
mdavis36: Not the CMake variable names but creating a CMake macro/function (we define some of them in `spheral/cmake/spheral/`) for switching out the library extensions to reduce code duplication in each tpl file.

src/DataBase/DataBase.cc
ptsuji: Ok, I've reverted this change.

mdavis36:
@ptsuji To get this moving we need two things:
1. To move the Communicator back to the Distributed package. Distributed has a sequential code path when MPI is disabled and should be used regardless of MPI status. 
2. The CMake macros / functions for Apple dynamic libraries.

ptsuji:
@mdavis36 I'm a little confused by (1), as currently the Distributed stuff is not built when MPI is disabled (there is a conditional on ENABLE_MPI in the src/Distributed/CMakeLists.txt). Are you saying we should remove that condition?

ldowen:
Hey @ptsuji I am working on this https://github.com/LLNL/spheral/pull/278 that should make things more clear. Once this is merged, you should be able to always include the Distributed source without issue.

jmikeowen:
Hey @ptsuji, I'm a bit confused as to why you guys were putting all those CPP guards around the calls to things like allReduce at all?  We were always building OK without MPI so far as I know, so why were those needed?

That said, like @ldowen mentioned, he's working on organizing our MPI vs. non-MPI stuff a bit more clearly, so once his changes are done hopefully whatever you guys were working around will be resolved.  We should talk more to understand what the problem was in the first place though.

ldowen:
I the issue is this line https://github.com/LLNL/spheral/blob/858faca549649a9b36888ad9b890e35ea90213a8/src/Distributed/CMakeLists.txt#L1 allowed the `Communicator.hh` to be found even when `ENABLE_MPI=OFF`. By leaving out the the Distributed directory entirely, you caused issues with `Communicator.hh` not being included at all.

jmikeowen:
But why wouldn't that tank our non-MPI build as well?

src/Utilities/globalNodeIDsInline.hh
ldowen, line 1: Looks like I missed some allreduce calls here in my own branch. I will fix that.

src/Mesh/MeshInline.hh
ptsuji: I've been merging Landon's branch (bugfix/central_mpi_calls) into mine to keep up-to-date.

cmake/tpl/polytope.cmake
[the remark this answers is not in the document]
ptsuji: I've created a function to handle the library extension on Mac.

ptsuji:
> @ptsuji To get this moving we need two things:
> 
> 1. To move the Communicator back to the Distributed package. Distributed has a sequential code path when MPI is disabled and should be used regardless of MPI status.
> 2. The CMake macros / functions for Apple dynamic libraries.

@mdavis36 I think I've addressed both of these questions, let me know if I need to do anything else.

cmake/InstallTPLs.cmake
```diff
@@ -85,6 +85,9 @@ endif()
 foreach(_comp ${AXOM_COMPONENTS_ENABLED})
   list(APPEND SPHERAL_BLT_DEPENDS axom::${_comp})
   get_target_property(axom_deps axom::${_comp} INTERFACE_LINK_LIBRARIES)
+  # strip cuda out so we have control over when cuda is enabled
+  list(REMOVE_DUPLICATES axom_deps)
+  list(REMOVE_ITEM axom_deps cuda)
```
mdavis36, line 90: I assume this is for when you want to build Axom with cuda but not spheral?

cmake/SetupSpheral.cmake
```diff
@@ -77,7 +77,7 @@ if(ENABLE_OPENMP)
   list(APPEND SPHERAL_BLT_DEPENDS openmp)
 endif()
 
-if(ENABLE_CUDA)
+if(ENABLE_CUDA AND NOT SPHERAL_HIDE_GPU)
```
mdavis36, line 80: Can you not set ENABLE_CUDA=off? is it because you're using Spheral as a submodule. Do we need to specialize this to something like SPHERAL_ENABLE_CUDA so you aren't having to throw SPHERAL_HIDE_GPU.

cmake/spheral/SpheralHandleExt.cmake
```diff
@@ -0,0 +1,32 @@
...
+#----------------------------------------------------------------------------------------
+
+function(Spheral_Handle_Ext lib_name libname APPLE)
+
+  if(APPLE)
+    set(SHARED_EXT "dylib")
+    set(${lib_name}_libs ${libname}.${SHARED_EXT})
+  endif()
+
+endfunction()
```
mdavis36, line 32: The documentation here doesn't match up with what this function is doing. This also seems to completely overwrites `${lib_name}_libs` to contain only a single library in the final list. I think a function that searches `${lib_name}_libs` for a set of libraries to be converted that finds and replaces them as necessary would be much more usefull.

cmake/tpl/polytope.cmake
```diff
@@ -1 +1,4 @@
 set(${lib_name}_libs libpolytope.a)
+
+include(${SPHERAL_ROOT_DIR}/cmake/spheral/SpheralHandleExt.cmake)
```
mdavis36, line 3: Includes for TPL functions can probably be moved to `cmake/InstallTPLs.cmake` so we don't need to keep redefining them.

mdavis36 (changes requested):
Some comments on the CMake changes here, may want to re-review this once #278 is merged and the diff cleans up a bit

cmake/SetupSpheral.cmake
mdavis36: I have changes that address this in one of the upcoming GPU PRs. Leave this how it is and we can discuss outside of this PR when it comes up again.

cmake/InstallTPLs.cmake
ptsuji: Yes, we build Axom with CUDA but not Spheral.

cmake/SetupSpheral.cmake
ptsuji: Ok, I'll leave this commented out here for now.

cmake/tpl/polytope.cmake
ptsuji: Ok, I've moved them to that file.

cmake/spheral/SpheralHandleExt.cmake
ptsuji: I've rewritten this function to replace .dylib or .so extensions with .a if ENABLE_STATIC_TPL=On.

cmake/SetupSpheral.cmake
```diff
@@ -77,6 +77,7 @@ if(ENABLE_OPENMP)
   list(APPEND SPHERAL_CXX_DEPENDS openmp)
 endif()
 
+#if(ENABLE_CUDA AND NOT SPHERAL_HIDE_GPU)
```
adayton1, line 80: Should this just be removed?

adayton1 (approved):
LGTM

cmake/SetupSpheral.cmake
mdavis36: If it's commented out and you dont need it right now then we can remove it. I was fine with this change as a temporary solution until we roll out the `SPHERAL_*` prefix changes that are coming. So either delete the comment or return it to using this logic.

mdavis36 (changes requested):
One last change on the CMake flag, after that I'm good with all the changes I can approve. We do need @jmikeowen to review this as well before merge.

cmake/SetupSpheral.cmake
ptsuji: Ok, I've removed it.

ptsuji:
@mdavis36 I believe I made the changes you requested for the CMake stuff, but I'm not sure how to change the status of that request (it still appears as "1 change requested" in the review). Can you resolve or update?

mdavis36 (dismissed):
I'm happy with CS related changes, but we still want @jmikeowen to look over this and approve before merging.

ptsuji:
@jmikeowen I've reverted the changes to make the "mask" variable automatically registered in the state. I've also changed some of the logic in computeFragmentField to look like computeVoronoiVolume, let me know if that looks good.

adayton1 (approved):
LGTM

src/Physics/GenericHydro.cc
jmikeowen, line 1: I know you mentined you changed this line due to time crash issues, but isn't this the wrong definition for the divergence in RZ now?

ptsuji: Ok, it looks like I was confused - if this is the trace of the strain tensor, then that last component added should be the velocity in the radial direction divided by the radial displacement. But I think it should be veli[1]*safeInv(posi[1]) then, instead of veli[0]*safeInv(posi[0]), since the y-axis corresponds to the r-axis in RZ space?

ptsuji: Ok, I've corrected it to be veli[1]*safeInv(posi[1]), and it runs fine on our test problems.

jmikeowen: Excellent -- good catch on this index error.

jmikeowen (approved):
Good to go again!

adayton1 (approved):
LGTM

PR #251, opened 2024-01-02
Setting more unique output test paths for planar porosity compaction

jmikeowen:
This PR is to make our CI testing with one of the new porosity tests (planar compaction) more robust.  It appears to have had some race conditions possible with removing directories when testing in parallel in the CI.

# Summary

- This PR is a bugfix
- It does the following:
  - Alters the output paths for one test when run under the ATS
  - Updates tolerances for several tests which occasionally were failing
  - Updates the local copy of lcats to run on the current LC machine configurations

------
### ToDo :

### DIFF
--- a/scripts/lc/lcats
+++ b/scripts/lc/lcats
@@ -23,7 +23,7 @@ def cpu_count():
     """
     if 'rzwhippet' in platform.node():
         return 56
-    elif 'rzgenie' in platform.node():
+    elif 'rzgenie' in platform.node() or 'ruby' in platform.node():
         return 36
     else:
         try:
--- a/scripts/lc/lcats
+++ b/scripts/lc/lcats
@@ -529,6 +529,23 @@ toss4BatchSettings= MachineInfo(
             ],
 )
 
+# Ruby settings (same as TOSS4 interactive without pdebug)
+rubySettings= MachineInfo(
+    name='toss4machine',
+    machineType='slurm36',
+    batch= False,
+    allocTime = 180,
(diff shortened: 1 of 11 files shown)


scripts/lc/lcats
```diff
@@ -506,16 +587,21 @@ elif 'glory' in SYS_TYPE:
 elif 'blue' in SYS_TYPE:
     machineSettings = blueosSettings
 
-elif 'toss' in SYS_TYPE:
-    if 'ruby' in LCSCHEDCLUSTER or 'rzgenie' in LCSCHEDCLUSTER or 'jadedev' == LCSCHEDCLUSTER or 'zindev' == LCSCHEDCLUSTER:
+elif 'toss_3' in SYS_TYPE:
+    if 'rzgenie' in LCSCHEDCLUSTER or 'jadedev' == LCSCHEDCLUSTER or 'zindev' == LCSCHEDCLUSTER:
```
mdavis36, line 591: should rzgenie be in the toss_4 condition now?

jmikeowen: rzgenie is running under the TOSS4 conditions now.  This particular branch is only executed if "toss_3" is in the SYS_TYPE, so it's skipped on rzgenie.  Now if for some reason they were to roll rzgenie back to TOSS3 this would activate again, in which case we might still want this check.  Really though we should just cull all the TOSS3 logic in the entire script once we're sure there are no more TOSS3 LC machines.

mdavis36 (approved):

PR #261, opened 2024-02-02
Bugfix/initialize problem startup dependencies

jmikeowen:
# Summary

- This PR is a bugfix (via an interface improvement)
- It does the following:
  - We have added a new optional initialization method for Physics packages: 
       Physics::initializeProblemStartupDependencies.
  - This optional hook is called once after Physics::initializeProblemStartup.
  - initializeProblemStartupDependencies is called with fully constructed State and StateDerivative objects, unlike initializeProblemStartup which is only called with a DataBase.
  - initializeProblemStartupDependencies is intended to be used to properly initialize dependent state (such as pressures and temperatures), which may depend on multiple physics packages and be non-trivial to update.
  - This corrects a bug whereby the ConstantBoundary could take a snapshot of incorrect initial state, freezing that incorrect state for all times thereafter.

------
### ToDo :

### DIFF
--- a/src/Physics/Physics.hh
+++ b/src/Physics/Physics.hh
@@ -100,7 +100,18 @@ public:
 
   //******************************************************************************//
   // An optional hook to initialize once when the problem is starting up.
+  // Typically this is used to size arrays once all the materials and NodeLists have
+  // been created.  It is assumed after this method has been called it is safe to
+  // call Physics::registerState for instance to create full populated State objects.
   virtual void initializeProblemStartup(DataBase<Dimension>& dataBase);
+
+  // A second optional method to be called on startup, after Physics::initializeProblemStartup has
+  // been called.
+  // One use for this hook is to fill in dependendent state using the State object, such as
+  // temperature or pressure.
+  virtual void initializeProblemStartupDependencies(DataBase<Dimension>& dataBase,
+                                                    State<Dimension>& state,
+                                                    StateDerivatives<Dimension>& derivs);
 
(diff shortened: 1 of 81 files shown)


src/Physics/Physics.hh
```diff
@@ -100,8 +100,19 @@ public:
...
   virtual void initializeProblemStartup(DataBase<Dimension>& dataBase);
 
+  // A second optional method to be called on startup, after Physics::initializeProblemStartup has
+  // been called.
+  // One use for this hook is to fill in dependendent state using the State object, such as
+  // temperature or pressure.
+  virtual void initializeProblemStartupDependencies(DataBase<Dimension>& dataBase,
+                                                    State<Dimension>& state,
+                                                    StateDerivatives<Dimension>& derivs);
+
```
brbass, line 115: These comments make more sense now that we have discussed them, but it would be nice to make it clearer what is expected at each point so new developers assume these are the **law of Mike** rather than suggestions:

- For `initializeProblemStartup`: "This is called after the materials and NodeLists are created. This method should set the sizes of all arrays owned by the physics package and initialize independent variables."
- For `initializeProblemStartupDependencies`: "This is called after independent variables have been initialized and put into the state and derivatives. During this method, the dependent state, such as temperature and pressure, is initialized so that all the fields in the initial state and derivatives objects are valid."

These are pretty similar to your current comments, so make changes at your own discretion!

brbass (approved):

jmpearl (approved):
looks good

mdavis36 (approved):

PR #271, opened 2024-03-28
Devops/docker envs

mdavis36:
# Summary

- This PR is a refactor
- It does the following:
  - Renames our `x86_64/packages.yaml` path to `Ubuntu20.04/packages.yaml` since that is the operating system the file is configured for.
  - Adds a `generic/` directory for external users to modify for their systems.
  - Documentation has been add outlining how to edit the `generic` config files for a system.
  - Uses `spack arch -o` to determine operating system and version string when`$SYS_TYPE` is not available

- Additional Changes:
  - Moves the installation location of python files and libraries to `<install_prefix>/lib/pythonX.Y/site-packages/Spheral`. This is standard practice for python library install structures. I Also ran into a bug where my filesystem on MacOS would confuse the `Spheral/` dir and the `spheral` executable file in our installation dir.
  - Added documentation for using docker dev environments with spheral for local development. 
  - Adding --fail-fast to spack install command to fail on first error.
  - Pull in changes from v2024.01.1 release that were never brought back to develop.

Docs: https://spheral.readthedocs.io/en/devops-docker-envs/index.html

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#81)


jmikeowen (approved):
Looking good

scripts/devtools/tpl-manager.py
```diff
@@ -149,12 +170,19 @@ def build_deps(args):
...
         print("** Building TPL's and generating host-config for {0}%{1} ...".format(package_name,s))
         os.environ["SPEC"] = s
         os.environ["LC_ALL"] = "en_US.UTF-8"
+
         if not args.no_spec:
             if sexe("{0} spec --fresh -I {1}@develop%{2}".format(spack_cmd, package_name, s), echo=True) : sys.exit(1)
-        if sexe("{0} dev-build --fresh --quiet --deprecated -u initconfig {2}@develop%{3} 2>&1 | tee -a \"dev-build-{3}-out.txt\"".format(spack_cmd, os.getcwd(), package_name, s), echo=True) : sys.exit(1)
+
+        # Install only the dependencies for Spheral
+        if sexe("{0} install --fail-fast --fresh --deprecated --only dependencies {2}@develop%{3} 2>&1 | tee -a \"dev-build-{3}-out.txt\"".format(spack_cmd, os.getcwd(), package_name, s), echo=True) : sys.exit(1)
```
ldowen, line 182: Why wasn't this install call needed before?

mdavis36: `dev-build` acts the same as `install` in regards to concretizing and installing dependencies, but `dev-build` doesn't have the `--fail-fast` option so we can have spack install only the dependencies with `install` and then let it create the `host-config` file from `dev-build` with our local repo. Using `install` tries to pull spheral develop and that isn't what we want for tpl-manager

PR #272, opened 2024-05-03
Update Spack to 0.22

ldowen:
# Summary

- This PR updates Spack and aspects of the TPLs
- It does the following:
  - The Spack hash used by uberenv is updated
  - The TPL Spack packages are updated or removed if possible
  - TPL manager logic is modified so only a single install call occurs
  - Spack configurations are updated
  - Package providers are set to avoid arbitrary concretization issues for downstream libraries

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#82)


cmake/tpl/zlib.cmake
```diff
@@ -1,5 +0,0 @@
-set(${lib_name}_libs libz.so)
-
-if(ENABLE_STATIC_TPL)
-  string(REPLACE ".so" ".a;" ${lib_name}_libs ${${lib_name}_libs})
-endif()
```
mdavis36, line 5: I'm pretty sure ALE3D use `EANBLE_STATIC_TPL`, is that still the case @ptsuji ?

scripts/devtools/tpl-manager.py
```diff
@@ -89,26 +89,13 @@ def parse_spec_list(file_path):
 #------------------------------
 # Dependencies
 #------------------------------
-def build_deps(args):
+def build_spack(args):
   print("~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~")
-  print("~~~~~ Building Dependencies")
+  print("~~~~~ Building and Configuring Spack")
```
mdavis36, line 94: We aren't necessarily building spack, just configuring it
```suggestion
  print("~~~~~ Configuring Spack")
```

scripts/devtools/tpl-manager.py
```diff
@@ -137,9 +124,9 @@ def build_deps(args):
     # We need to install spack without any configuration files so we can use
     # spack arch to determine the OS of the system and later to use spack find
     # for generating external package files on external systems.
-    if sexe("python3 {0} --setup-only {1} {2} {3} {4}".format(uberenv_path, prefix_opt, uberenv_project_json_opt, spack_upstream_opt, uberenv_spack_url_opt), echo=True): sys.exit(1)
+    sexe("python3 {0} --setup-only {1} {2} {3} {4}".format(uberenv_path, prefix_opt, uberenv_project_json_opt, spack_upstream_opt, uberenv_spack_url_opt))
```
mdavis36, line 127: What happens if Uberenv fails now?

scripts/devtools/tpl-manager.py
```diff
@@ -150,12 +137,30 @@ def build_deps(args):
 
 
   # Setup spack w/ Uberenv and the appropriate external package/compiler configs.
-  if sexe("python3 {0} --setup-only {1} {2} {3} {4} {5}".format(uberenv_path, prefix_opt, uberenv_project_json_opt, spack_config_dir_opt, spack_upstream_opt, uberenv_spack_url_opt), echo=True): sys.exit(1)
+  sexe("python3 {0} --setup-only {1} {2} {3} {4} {5}".format(uberenv_path, prefix_opt, uberenv_project_json_opt, spack_config_dir_opt, spack_upstream_opt, uberenv_spack_url_opt))
```
mdavis36, line 140: Again, what if this fails?

scripts/devtools/tpl-manager.py
```diff
@@ -150,12 +137,30 @@ def build_deps(args):
...
+  if args.spec_list:
+    spec_list = parse_spec_list(args.spec_list)
+  elif args.spec:
+    spec_list.append(args.spec)
+  else:
+    print("ERROR: Please define --spec or --spec-list, aborting...")
+    sys.exit(1)
+  for s in spec_list:
+    print("** SPEC : {0}".format(s))
+  spack_cmd=os.path.join(args.spheral_spack_dir, "spack/bin/spack")
```
mdavis36, line 163:
```suggestion
  spack_cmd=os.path.join(args.spheral_spack_dir, "spack/bin/spack")
  
```

scripts/devtools/tpl-manager.py
```diff
@@ -150,12 +137,30 @@ def build_deps(args):
...
   if os.path.exists(spheral_config_dir+"/concretizer.yaml"):
-    sexe("cp {0}/concretizer.yaml {1}".format(spheral_config_dir, os.path.join(args.spheral_spack_dir, "spack/etc/spack/defaults")), echo=True)
+    sexe("cp {0}/concretizer.yaml {1}".format(spheral_config_dir, os.path.join(args.spheral_spack_dir, "spack/etc/spack/defaults")))
 
+def build_deps(args):
+  print("~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~")
+  print("~~~~~ Building Dependencies")
+  print("~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~")
+  print("")
+  print("{0}".format(project_dir))
```
mdavis36, line 151:
```suggestion
  print("{0}".format(project_dir))
  
```

scripts/devtools/tpl-manager.py
```diff
@@ -170,27 +175,25 @@ def build_deps(args):
...
-            if sexe("{0} spec --fresh -I {1}@develop%{2}".format(spack_cmd, package_name, s), echo=True) : sys.exit(1)
-
-        # Install only the dependencies for Spheral
-        if sexe("{0} install --fail-fast --fresh --deprecated --only dependencies {2}@develop%{3} 2>&1 | tee -a \"dev-build-{3}-out.txt\"".format(spack_cmd, os.getcwd(), package_name, s), echo=True) : sys.exit(1)
+          sexe("{0} spec --fresh -IL {1}@develop%{2} 2>&1 | tee -a \"spec-info-{2}-out.txt\"".format(spack_cmd, package_name, s))
 
-        # Using dev-build we can have spack generate an init-config with the local source files for spheral.
-        if sexe("{0} dev-build --fresh --deprecated -u initconfig {2}@develop%{3} 2>&1 | tee -a \"dev-build-{3}-out.txt\"".format(spack_cmd, os.getcwd(), package_name, s), echo=True) : sys.exit(1)
+        # Install only the dependencies for Spheral and create CMake configure file
+        sexe("{0} install --fail-fast --fresh -u initconfig {1}@develop%{2} dev_path=./ 2>&1 | tee -a \"tpl-build-{2}-out.txt\"".format(spack_cmd, package_name, s))
```
mdavis36, line 186: Does `dev_path` have any other side effects, or alter paths for how the dependencies are being built? Or does it only effect the spec/package you are targeting (so spheral only in this case)?

scripts/spack/configs/blueos_3_ppc64le_ib/packages.yaml
```diff
@@ -2,13 +2,38 @@ packages:
   all:
     target: [ppc64le]
     compiler: [gcc, clang]
+    providers:
+      awk: [gawk]
+      blas: [openblas, amdblis]
```
mdavis36, line 7: Are we depending on awk or blas anywhere?

scripts/spack/configs/config.yaml
```diff
@@ -71,18 +99,139 @@ config:
...
+  url_fetch_method: urllib
+
+  # The maximum number of jobs to use for the build system (e.g. `make`), when
+  # the -j flag is not given on the command line. Defaults to 16 when not set.
+  # Note that the maximum number of jobs is limited by the number of cores
+  # available, taking thread affinity into account when supported. For instance:
+  # - With `build_jobs: 16` and 4 cores available `spack install` will run `make -j4`
+  # - With `build_jobs: 16` and 32 cores available `spack install` will run `make -j16`
+  # - With `build_jobs: 2` and 4 cores available `spack install -j6` will run `make -j6`
+  # build_jobs: 16
```
mdavis36, line 159: We've lost build_jobs: 54 here, we might want to increase this again

scripts/spack/configs/toss_4_x86_64_ib/packages.yaml
```diff
@@ -4,12 +4,37 @@ packages:
     # us to run on broadwell as well
     target: [ivybridge]
     compiler: [gcc, clang]
+    providers:
+      awk: [gawk]
+      blas: [openblas, amdblis]
```
mdavis36, line 9: Same as above

scripts/spack/packages/opensubdiv/package.py
```diff
@@ -28,15 +28,15 @@ def url_for_version(self, version):
...
 
     depends_on('cmake@2.8.6:', type='build')
-    depends_on('graphviz', type='build', when='+doc')
-    depends_on('doxygen', type='build', when='+doc')
-    depends_on('py-docutils', type='build', when='+doc')
-    depends_on('python@2.6:2', type='build', when='+doc')
+    # depends_on('graphviz', type='build', when='+doc')
+    # depends_on('doxygen', type='build', when='+doc')
+    # depends_on('py-docutils', type='build', when='+doc')
+    # depends_on('python@2.6:2', type='build', when='+doc')
```
mdavis36, line 39: Is this a change from the latest package file, or do we need to set `~doc` on our opensubdiv dependency?

scripts/spack/packages/spheral/package.py
```diff
@@ -203,3 +197,14 @@ def cmake_args(self):
...
+
+    @property
+    def build_dirname(self):
+        """Directory name to use when building the package."""
+        return "spack-build-%s" % self.pkg.spec.dag_hash(7)
+
+    @property
+    def build_directory(self):
+        """Full-path to the directory to use when building the package."""
+        return os.path.join(self.pkg.stage.path, self.build_dirname)
```
mdavis36, line 209: Is this to get `dev_path` to work what is this?

mdavis36 (changes requested):

scripts/spack/packages/opensubdiv/package.py
ldowen: This is from the latest package file.

scripts/spack/packages/spheral/package.py
```diff
@@ -49,35 +49,32 @@ class Spheral(CachedCMakePackage, CudaPackage):
...
+    depends_on('py-numpy-stl', type='build')
+    depends_on('py-matplotlib backend=tkagg +fonts', type='build')
+    depends_on('py-h5py', type='build')
+    depends_on('py-docutils', type='build')
+    depends_on('py-scipy', type='build')
+    depends_on('py-ats', type='build')
+    depends_on('py-mpi4py', type='build', when='+mpi')
+
+    depends_on('py-sphinx', type='build')
+    depends_on('py-sphinx-rtd-theme', type='build')
```
mdavis36, line 77: We want to lock these down to working versions for reproducability across systems

scripts/spack/packages/spheral/package.py
ldowen: This is in preparation to moving to spack environments. Forgot I left it in there. I can get rid of it but it shouldn't make a difference.

scripts/spack/packages/spheral/package.py
```diff
@@ -36,11 +36,11 @@ class Spheral(CachedCMakePackage, CudaPackage):
...
     depends_on('cmake@3.10.0:', type='build')
 
-    depends_on('zlib@1.2.11 +shared +pic', type='build')
-
     depends_on('boost@1.74.0 +system +filesystem -atomic -container -coroutine -chrono -context -date_time -exception -fiber -graph -iostreams -locale -log -math -mpi -program_options -python -random -regex -test -thread -timer -wave +pic', type='build')
 
-    depends_on('qhull@2020.1 +pic', type='build')
+    depends_on('zlib@1.3 +shared +pic', type='build')
+
+    depends_on('qhull +pic', type='build')
```
mdavis36, line 43: Needs a version

scripts/spack/packages/spheral/package.py
mdavis36: We can leave it in if you can tell me what `dag_hash(7)` means

scripts/devtools/tpl-manager.py
ldowen: The `dev_path` stuff is essentially what the dev-build command is doing under the hood. It shouldn't do anything different regarding building the TPLs except it does provide a symbolic link to the spheral build directory for some reason.

cmake/tpl/zlib.cmake
mdavis36: Maybe @brbass is using this too?

scripts/devtools/tpl-manager.py
ldowen: The `check=True` should cause the tpl manager to fail. Unfortunately, if something fails during a spack command, the current and previous methods failed to cause the tpl manager to quit running. But if uberenv exits with a fail flag, it should stop properly.

mdavis36: I'm more concerned with ubuerenv failing to find a spack repository on an air-gapped system where spack needs to be local

scripts/spack/configs/blueos_3_ppc64le_ib/packages.yaml
```diff
@@ -2,15 +2,44 @@ packages:
...
 
   cuda:
-    version: [11.1.0, 11.0.2, 10.1.243, 10.1.168, 9.2.148, 8.0]
+    version:
+    - 11.1.0
+    - 11.0.2
+    - 10.1.243
+    - 10.1.168
+    - 9.2.148
+    - '8.0'
```
ldowen, line 42: @mdavis36 I'm not sure why spack put 8.0 in quotes here.

scripts/spack/packages/spheral/package.py
mdavis36: Sounds good

cmake/tpl/zlib.cmake
ptsuji: Yes, we're using ENABLE_STATIC_TPL.

brbass: We are also using it.

cmake/SetupSpheral.cmake
```diff
@@ -83,8 +82,6 @@ if(ENABLE_CUDA)
   list(APPEND SPHERAL_CXX_DEPENDS cuda)
 endif()
 
-option(BOOST_HEADER_ONLY "only use the header only components of Boost" OFF)
```
ptsuji, line 86: We use the header-only boost option, is it being removed because it's the default?

cmake/InstallTPLs.cmake
```diff
@@ -89,7 +89,7 @@ foreach(_comp ${AXOM_COMPONENTS_ENABLED})
 endforeach()
 
 # TPLs that must be imported
-list(APPEND SPHERAL_EXTERN_LIBS zlib boost eigen qhull silo hdf5 polytope)
+list(APPEND SPHERAL_EXTERN_LIBS boost eigen qhull silo hdf5 polytope)
```
ptsuji, line 92: We build our own zlib to link to, is this being removed because the system zlib will be used?

cmake/SetupSpheral.cmake
ldowen: It was thought to be unnecessary since we no longer build boost with cmake. Does it actually make a difference for your builds?

cmake/InstallTPLs.cmake
ldowen: The zlib path will be included in the CMAKE_PREFIX_PATH in the cmake file created when running TPL. But the zlib target isn't explicitly used when building Spheral.

scripts/spack/packages/spheral/package.py
```diff
@@ -34,13 +34,13 @@ class Spheral(CachedCMakePackage, CudaPackage):
     # DEPENDS
     # -------------------------------------------------------------------------
     depends_on('mpi', when='+mpi')
-    depends_on('cmake@3.10.0:', type='build')
-
-    depends_on('zlib@1.2.11 +shared +pic', type='build')
+    depends_on('cmake@3.18.0:', type='build')
 
     depends_on('boost@1.74.0 +system +filesystem -atomic -container -coroutine -chrono -context -date_time -exception -fiber -graph -iostreams -locale -log -math -mpi -program_options -python -random -regex -test -thread -timer -wave +pic', type='build')
```
ptsuji, line 39: Will the Boost version be updated in this branch?

ldowen: No. Is that something you need updated?

cmake/SetupSpheral.cmake
mdavis36: Were hoping to do a separate PR soon where we can update TPLs to more recent versions. This PR only updates TPLs in order to make the latest version of spack happy.

scripts/spack/packages/spheral/package.py
mdavis36: Were hoping to do a separate PR soon where we can update TPLs to more recent versions. This PR only updates TPLs in order to make the latest version of spack happy.

.gitlab/os.yml
```diff
@@ -4,7 +4,7 @@
 .sys_config:
   variables:
     LLNL_SERVICE_USER: sphapp
-    UPSTREAM_DIR: /usr/WS2/wciuser/Spheral/spheral-spack-tpls
+    UPSTREAM_DIR: /usr/WS2/sduser/Spheral/0.22
```
mdavis36, line 7: Can we name the directory something a little more descriptive than 0.22 so we know its the spack tpl upstream directory.

scripts/spack/configs/ubuntu20.04/packages.yaml
```diff
@@ -4,13 +4,26 @@ packages:
...
     buildable: false
-
+  all:
+    variants: +pic
+    providers:
+      mpi: [mpich]
+      blas: [netlib-lapack]
+      lapack: [netlib-lapack]
+      pkgconfig: [pkg-config]
+      pkgconf: [pkg-config]
```
mdavis36, line 14: Do we need both of these?

scripts/spack/configs/ubuntu20.04/packages.yaml
```diff
@@ -4,13 +4,26 @@ packages:
     - spec: mpich@3.3.2
       prefix: /usr
     buildable: false
-
+  all:
+    variants: +pic
```
mdavis36, line 8: Why does ubuntu need this?

mdavis36 (changes requested):
Good work on getting through this slog, hopefully we can start to update more regularly and avoid such a heavy lift in the future. Just a couple of questions and one change request in terms of our upstream directory name. I hope it isn't too much of a pain to move ...

.gitlab/os.yml
ldowen: Yeah let me look into how I can do that without having to reinstall everything in the upstream. It should be doable.

scripts/spack/configs/ubuntu20.04/packages.yaml
ldowen: I think so. I don't think pkgconfig is a virtual package so we have to be specific. I could be wrong though.

scripts/spack/configs/ubuntu20.04/packages.yaml
ldowen: Not sure if it does. It is just there in case the variant exists for a library and we need to make it shared. I wasn't sure if there are some system libraries that get installed on ubuntu that we use external versions for on LC machines. I can get rid of it and see if it still works.

cmake/tpl/zlib.cmake
ldowen: We don't need to explicitly bring zlib in as a target so we don't use this file anymore. The zlib tpl is included in the `CMAKE_PREFIX_PATH` instead.

mdavis36 (approved):

mdavis36:
@ldowen LGTM

PR #278, opened 2024-06-26
Cleanup MPI calls

ldowen:
# Summary
This PR improves the interface to the MPI functionality.

- This PR is a refactoring and bugfix
- It does the following:
  - Refactors the `allReduce` and `scan` wrappers to use the `Communicator::communicator()` by default and puts both in a single file.
  - Refactors the `Distributed` source directory, having it provide the MPI wrappers even when MPI is not enabled; this prevents potential bugs from occurring when wrappers like `Communicator` are present but `Communicator.hh` is not available.
  - Replaces the `MPI_SUM`, `MPI_MAX`, etc with `SPHERAL_OP_SUM`, `SPHERAL_OP_MAX` etc

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#88)


mdavis36:
What do you guys think about using something like `SPHERAL_OP_MIN`, `SPHERAL_OP_MAX` etc. Instead of `SPHERAL_MPI_MIN`, `SPHERAL_MPI_MAX`. It feels like those macros suggest an MPI call instead of just defining an execution context agnostic operator.

ldowen:
Yeah I see what you mean. That works for me. I will change it now.

src/Distributed/allReduce.hh
```diff
@@ -0,0 +1,79 @@
...
+
+template<typename Value>
+constexpr Value
+allReduce(const Value& value, const int /*op*/, const int comm = 0) {
+  return value;
+}
+
+template<typename Value>
+constexpr Value
+scan(const Value& value, const int /*op*/, const int comm = 0) {
```
mdavis36, line 72: This should throw a warning but I think we have them turned off, we don't need `comm` here as it isn't used, just the type `const int`.

src/FractalStruct/check_for_edge_trouble.cc
```diff
@@ -1,46 +1,46 @@
...
+	if(pos[2] >= 1.0) 
+	  pos[2]-=eps;
+	else if(pos[2] <= 0.0) 
+	  pos[2]+=eps;
+	p->set_pos(pos);
+      }
+    FileFractal << " Total Outsiders " << outsiders << "\n";
+    fractal.timing(1,2);
+  }
+}
```
mdavis36, line 46: What happened here? I can't see any difference

src/KernelIntegrator/FlatConnectivity.cc
```diff
@@ -329,17 +326,12 @@ computeGlobalIndices(const DataBase<Dimension>& dataBase,
   VERIFY(numInternalNodesDB == mNumInternalLocalNodes);
   
   // Get global indices manually
-#ifdef USE_MPI
-  int globalScan = 0;
-  MPI_Scan(&mNumInternalLocalNodes, &globalScan, 1, MPI_INT, MPI_SUM, Communicator::communicator());
+  int globalScan = scan(mNumInternalLocalNodes, SPHERAL_OP_SUM);
```
mdavis36, line 329: `scan()` looks a whole lot like it could be a standard library function call, and at the very least it's not entirely obvious this could be executed in a distributed manner. We might want to either rename this, maybe `distScan`, or namespace the functions so `distributed::scan` & `distributed::allReduce`.

mdavis36 (changes requested):

src/FractalStruct/check_for_edge_trouble.cc
ldowen: My python script to replace the function names must have changed it somehow. I will undo it.

mdavis36 (dismissed):

src/Damage/IvanoviSALEDamageModel.cc
jmikeowen, line 1: If this include isn't used here should it just be removed?

ldowen: Yep. Good catch

mdavis36 (approved):

PR #283, opened 2024-07-08
Hotfix CD pipeline

mdavis36:
# Summary

- This PR is a release hotfix and will be tagged as **v2024.06.1** after merge...
- It does the following:
  - Moves the steps for creating our dev-pkg / tar file out of gitlab and into a standalone script.
  - Fixes the broken Continuous Deployment pipeline that was silently broken during the latest spack update.
    - The `-d` option for `spack buildcache` is no longer supported. We need to first create a mirror and then point the cache to that mirror.
  - Possibly fixes a bug seen when building on air gapped builds where concretization is different from the buildcache and the final install.
    - Flags for concretization and installation need to be the same in this case we are setting `--fresh` and `--deprecated`.
  - Sets `libjpeg` as the provider package for the virtual spack package `jpeg` for toss4 builds.
    - During development it was noticed that sometimes `py-pillow` would try to build with providers of `jpeg` that were not the basic `libjpeg` package during the installation locking this seems to resolve the issue.

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#90)


jmikeowen:
If we're doing a patch release could we go ahead and bring in that fix for our use of Axom in the polyhedron containment query?  Even though the error is probably harmless, might as well fix this known issue if we're doing a patch release.

ldowen (approved):

PR #289, opened 2024-07-23
Value-View Interface

mdavis36:
# Summary

- This PR is a (refactoring, bugfix, feature, something else)
- It does the following:
  - Modifies / refactors (class or method) (how?)
  - Fixes (issue number(s))
  - Adds (specific feature) at the request of (project or person)

------
### ToDo :


cmake/SetupSpheral.cmake
```diff
@@ -140,6 +148,10 @@ set_property(GLOBAL PROPERTY SPHERAL_CXX_DEPENDS "${SPHERAL_CXX_DEPENDS}")
 #-------------------------------------------------------------------------------
 # Prepare to build the src
 #-------------------------------------------------------------------------------
+configure_file(${SPHERAL_ROOT_DIR}/src/config.hh.in
```
ldowen, line 151: I think this is happening twice now? Once here and once in src/CMakeLists.txt

mdavis36: I moved it here. I don't see it in `src/CMakeLists.txt` anymore.

ldowen: https://github.com/LLNL/spheral/blob/d4cada370c1b4dcf8b225ca0a775b300548ba18f/src/CMakeLists.txt#L75
Isn't this it?

cmake/InstallTPLs.cmake
```diff
@@ -86,11 +86,13 @@ message("-----------------------------------------------------------------------
 find_package(RAJA REQUIRED NO_DEFAULT_PATH PATHS ${raja_DIR})
 if (RAJA_FOUND) 
   message("Found RAJA External Package.")
+  blt_convert_to_system_includes(TARGET RAJA)
 endif()
 message("-----------------------------------------------------------------------------")
 find_package(umpire REQUIRED NO_DEFAULT_PATH PATHS ${umpire_DIR})
 if (umpire_FOUND) 
   message("Found umpire External Package.")
+  blt_convert_to_system_includes(TARGET umpire)
```
ptsuji, line 95: Should this blt_convert_to_system_includes() stuff also be in the block under "Found chai External Package"?

cmake/SetupSpheral.cmake
```diff
@@ -85,7 +90,10 @@ if(ENABLE_CUDA)
   #set(CMAKE_CUDA_FLAGS  "${CMAKE_CUDA_FLAGS} -arch=${CUDA_ARCH} --expt-relaxed-constexpr --extended-lambda -Xcudafe --display_error_number")
   set(CMAKE_CUDA_STANDARD 17)
   list(APPEND SPHERAL_CXX_DEPENDS cuda)
+  set(SPHERAL_ENABLE_CUDA On)
+  set(SPHERAL_ENABLE_VVI On)
```
ptsuji, line 94: Should the Value-View interface be enabled for CPU builds as well?

src/CMakeLists.txt
```diff
@@ -80,6 +79,8 @@ foreach(_package ${_packages})
   add_subdirectory(${_package})
 endforeach()
 
+#blt_print_target_properties(TARGET Spheral_Utilities)
```
ptsuji, line 82: Could probably remove this comment.

cmake/SetupSpheral.cmake
mdavis36: It *can* be used for CPU builds but it is *required* for any GPU builds where we will be executing on the device. This just ensures it's enabled if you are building spheral w/ cuda.

cmake/InstallTPLs.cmake
mdavis36: I`ll add it to CHAI. We already do this for all other external packages on line 127.

cmake/SetupSpheral.cmake
mdavis36: Yep, that change wasn't committed, good catch

cmake/InstallTPLs.cmake
ldowen: We don't do it for the packages we use `find_package` on. We only do it on packages we manually bring in. Was this necessary for your stuff?

mdavis36: It was, pedantic warnings from RAJA were causing our WARNINGS_AS_ERRORS builds to fail.

cmake/SetupSpheral.cmake
```diff
@@ -153,6 +165,7 @@ endif()
 # Build C++ tests and install tests to install directory
 #-------------------------------------------------------------------------------
 if (ENABLE_TESTS)
+  add_subdirectory(${SPHERAL_ROOT_DIR}/tests)
   add_subdirectory(${SPHERAL_ROOT_DIR}/tests/unit/CXXTests)
 
   # A macro to preserve directory structure when installing files
```
ldowen, line 171: Do we want the `install_with_directory` command to also find the `tests/cpp` or should those be left out of that command?

jmikeowen (approved):
Looking good once tests all pass

cmake/SetupSpheral.cmake
mdavis36: Should we address this in #304

ldowen: Sure that works for me.

ldowen (dismissed):

src/Kernel/TableKernel.cc
```diff
@@ -184,9 +208,9 @@ TableKernel<Dimension>::TableKernel(const KernelType& kernel,
   mNumPoints(numPoints),
   mMinNperh(std::max(minNperh, 1.1/kernel.kernelExtent())),
   mMaxNperh(maxNperh),
-  mInterp(0.0, kernel.kernelExtent(), numPoints,      [&](const double x) { return kernel(x, 1.0); }),
-  mGradInterp(0.0, kernel.kernelExtent(), numPoints,  [&](const double x) { return kernel.grad(x, 1.0); }),
-  mGrad2Interp(0.0, kernel.kernelExtent(), numPoints, [&](const double x) { return kernel.grad2(x, 1.0); }),
+  mInterp(0.0, kernel.kernelExtent(), numPoints, Wlookup<KernelType>(kernel)),
+  mGradInterp(0.0, kernel.kernelExtent(), numPoints, gradWlookup<KernelType>(kernel)),
+  mGrad2Interp(0.0, kernel.kernelExtent(), numPoints, grad2Wlookup<KernelType>(kernel)),
```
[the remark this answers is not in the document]
jmikeowen: Ah I see, so you have to go back to externally defined functors.  This is harmless enough, but kind of a bummer since I like the compact (and I think clearer) lambda function version.  Do we think this is something that newer versions of nvcc can/will correct?  Seems like they're not supporting standard C++ things here.

Also, since RAJA relies on lamda functions for it's trickery as well, what's teh salient difference here?  Is it the reference capturing [&] version?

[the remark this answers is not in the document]
mdavis36: RAJA requires capturing by value [=] so a copy of the object must be taken into a host-device lambda.

In this specific case the lambda is used for construction of the Interpreter. When VVI + Cuda are enabled the underlying ManagedSharedPtr tries to invoke the constructor on both the host and device. Meaning `kernel` isn't available to the lambda invoked on the device, we need a copy of it to be passed over to the device in order to be able to call it correctly. 

There are a couple of ways of getting back to an interface where we provide a lambda:
1. Construct interpreters from factory functions that are able to construct w/ appropriate size on Host + Device, then fill the interpreter on the host and copy the contents to the device.
2. Have the interprter take a handle to the kernel object then define the lambda as `[=](const double, const kernel) { return kernel ... }` so the interpreter can pass the kernel internally in it's constructor.

PR #297, opened 2024-08-13
Make FSISPH package optional

### LINKED ISSUE #290: Allow FSISPH package to be disabled
As far as I can tell, FSISPH is self-contained except that it bleeds into one file in Hydro (PressurePolicy.cc). Is it possible to make FSISPH depend on Hydro instead of the other way around so that FSISPH can be disabled?

adayton1:
# Summary

- This PR is a feature
- It does the following:
  - Moves the damaged pressure field name to SolidFieldNames so that both Hydro and FSISPH can use it
  - Makes the FSISPH package optional to reduce build times and library size
  - Filter out FSISPH tests when FSISPH is disabled
  - Fixes #290 

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR# https://rzlc.llnl.gov/gitlab/owen/LLNLSpheral/-/merge_requests/100)


jmpearl:
I'm a little concerned about the damaged pressure field name being added to the HydroFieldNames. The damaged pressure field list is relate to the solid material response and it make by more appropriate to add it to the SolidFieldNames if it need to be migrated out of the FSI dir. 

The other issue I forsee, there are now two versions of the damaged pressure field name, one located in HydroFieldNames and one in FSIFieldNames. If one of these is altered down the road it could lead to bugs that are difficult to track down. However we decide to handle this I think the damaged pressure should only be attached to one fieldNames file.

jmikeowen:
This mostly looks fine to me.  I agree with @jmpearl on the damagedPressure field name though.  We should move it to the SolidFieldNames, and there definitely should only be one copy of this name.  So I guess remove it from FSIFieldNames.

adayton1:
Thanks for your suggestions! I've moved the damaged pressure field name to SolidFieldNames, which is then used in both Hydro and FSISPH.

jmpearl:
This looks good to me, what does the integration.ats edit do?

glue(fsisph = False)
glue(gsph = False)
glue(svph = False)

jmpearl (approved):

adayton1:
> This looks good to me, what does the integration.ats edit do?
> 
> glue(fsisph = False) glue(gsph = False) glue(svph = False)

This tells ATS that by default the tests are not using these packages. You'll notice in the tests that do use them, we explicitly set these values to True. I had to do this so that ATS would filter out these tests when the package is turned off.

ptsuji (approved):

PR #299, opened 2024-08-19
Refactor explicit instantiations

adayton1:
# Summary

- This PR is a refactoring
- It does the following:
  - Uses a CMake generated config header to control 1D, 2D, and 3D instantiations
  - Avoids using python to generate explicit instantiations - they can be added directly to a .cc file

------

### ToDo :


jmikeowen:
Now this change I think we need to have a larger conversation about -- not sure I want to go this way for the instantiations.  Don't get too far working on this 'til we have a chance to discuss with myself, Mike (Davis), and Landon.

adayton1:
These changes actually slowed the shared library build down. It looks like the biggest portion of time is taken while parsing axom, fmt in particular. It looks like axom is brought in virtually everywhere because Dimension.hh includes GeomPolyhedron.hh, which includes axom. I think separating out GeomPolyhedron.hh from Dimension.hh would result in the biggest build time savings.

jmikeowen:
It might make more sense to move the axom include to the GeomPolyhedron.cc (along with any implementations that require it).

From: Alan Dayton ***@***.***>
Date: Wednesday, August 28, 2024 at 8:05 AM
To: LLNL/spheral ***@***.***>
Cc: Mike Owen ***@***.***>, Comment ***@***.***>
Subject: Re: [LLNL/spheral] Refactor explicit instantiations (PR #299)

These changes actually slowed the shared library build down. It looks like the biggest portion of time is taken while parsing axom, fmt in particular. It looks like axom is brought in virtually everywhere because Dimension.hh includes GeomPolyhedron.hh, which includes axom. I think separating out GeomPolyhedron.hh from Dimension.hh would result in the biggest build time savings.

adayton1:
Good point! I'll see if the axom include can be moved to the .cc file

PR #302, opened 2024-08-22
New ASPH H update algorithm

jmikeowen:
# Summary

- This PR is a feature updating how we update the smoothing scales
- It does the following:
  - Introduces a new method for updating the ASPH ideal H
  - Splits the H update algorithms into their own physics packages, removing them (and their needed data) from all hydro packages.
  - Updates how the Voronoi-like geometry is computed, making it its own service physics package akin to how the RK corrections package is handled.
  - Introduces a new experimental general hourglass control for all hydro algorithms.
  - Cleans up all Clang compiler warnings, and sets the Clang CI jobs to use Warnings as Errors to keep those builds warning free.

------
### ToDo :


ptsuji:
@jmikeowen when I built this branch with SpheralC tests, the Taylor impact tests will croak in the call to spheral_initialize_boundaries_and_physics() (or initializeBoundariesAndPhysics() in SpheralPseudoScript). It seems the DataBase pointer is trying to reinitialize neighbors and getting the smoothing kernel extent runs into a segfault.

jmikeowen:
I've changed the interfaces to several hydro packages, so SpheralC will
have to be updated.

src/Integrator/Integrator.cc
```diff
@@ -389,7 +391,9 @@ void
 Integrator<Dimension>::
 appendPhysicsPackage(Physics<Dimension>& package) {
   if (!havePhysicsPackage(package)) {
+    for (auto* packagePtr: package.preSubPackages()) this->appendPhysicsPackage(*packagePtr);
     mPhysicsPackages.push_back(&package);
+    for (auto* packagePtr: package.postSubPackages()) this->appendPhysicsPackage(*packagePtr);
```
brbass, line 396: It looks like the pre/postSubPackages are only used for the smoothing scale at the moment. A few questions:

- Are there other examples we should consider of things that could become subpackages? 
- Is there a way to do this without the extra complexity of subpackages? If we were missing an appendBoundary instance for one of the subpackages, for instance, that could be difficult to test and find. Although the RK and Voronoi packages are is added directly to the integrator without the extra subpackage indirection, those aren't specific to the hydro package, so maybe this extra machinery makes sense. 
- Do we need to be worried about the case where multiple packages include the same subpackage? If so, we may want to add a check for `if (!havePhysicsPackage(subPackage))`. It seems like the use case is that a subpackage will never be tied to a different package.

jmikeowen: Those are good questions.  I added these subpackages as a workaround so we could still maintain the user script level ideas of SPH/ASPH, FSISPH/AFSISPH, etc.  I don't know of a better way to maintain that sort of script level backwards compatibility, and I think those concepts of the hydro including the H evolution makes sense from a user perspective.

Another example of something that might become a subpackage is the artificial viscosities.  We could abstract those out of the hydro packages as well, particularly if we add the idea of pair-wise centering data type.

At the moment I don't forsee having subpackages that might be defined by multiple packages, but we can easily check for that since the way the work is the SpheralController builds the ultimate package list by adding in the subpackages from each package.

src/Hydro/SpecificThermalEnergyPolicy.cc
jmpearl, line 1: I think we also need the self interaction block in the CompatibleDifferenceSpecificThermalEnergyPolicy.cc for FSISPH, MFM, MFV and GSPH to conserve energy properly w/ the hour glass acceleration?

src/SPH/PSPHHydroBase.cc
jmpearl, line 1: It looks like the applyFieldListGhostBoundary() methods are applied twice here: once in the postStateUpdate method of PSPH then again when it returns true to the integrator?

jmpearl:
All the edits to FSISPH, MFM, MFV, and GSPH look good to me. I didn't get through all the files but I gave most of the other hydros a quick eye too.

src/Hydro/SpecificThermalEnergyPolicy.cc
jmikeowen: The current SubPointPressureHourglassControl does not create self-interactions (though at one stage it did).  However, you are right that for consistency between these various compatible energy update policies we should handle the possible existence of self-interaction terms similarly.  Really we need to look at merging these all into one policy, but for now I'll introduce the same self-interaction check in the CompatibleDifferenceSpecificThermalEnergyPolicy.

jmikeowen: Oh shoot, there is a problem.  That policy also needs the pairDepsDt array to be properly sized, which it will not be (I don't think) if we're allowing self-interactions.  Since there are no main-line algorithms that currently have self-interactions I say we leave this as is.  We'll have to revisit this though if we for instance introduce an RZ version of any of those hydros (FSI, MFM, MFV, GSPH) since those do have self-interaction terms.

jmpearl (dismissed):

ldowen (approved):

PR #308, opened 2024-10-08
bugfix FSISPH and constant boundaries

jmpearl:
# Summary

- the damaged pressure was being initialized at 2x. This would cause issues for problems w/ constant boundaries.

------
### ToDo :


jmikeowen (approved):
Assuming the damaged pressure is always otherwise initialized this is fine.  Another fix would have been to initialize here with 
mDamagedPressure = this->pressure();
rather than adding it.  Why was this done with += originally?

jmpearl:
That is a great question and i asked myself the same thing. I can't remember why I had it as plus equals originally. I think the damaged pressure should always be initialized now through the call to the pressure update policy. Do you see any edge cases where that might not be the case?

jmikeowen:
Naw, I think you should be good as is.  Still have to ask about having a test that would catch this error of course... :-)

jmpearl:
I'm going to merge this now just to get the change in and punt on the test to a later date

PR #314, opened 2024-11-08
Pip Managed Runtime Dependencies

mdavis36:
# Summary

- This PR is a refactoring of our package management system for python packages in the Spheral build pipeline.

### Spheral +python variant
- The spheral spack package will default to `+python`. This will add python as a dependency for building spheral. It will also enable `+python` for polytope.
- We no longer need any of our pip package dependencies in spacke. This significantly cuts down our build tree.

#### Build Time Deps
- We still need spack to build python for us :( . 
  - The defaults on LC fails to build `mpi4py` correctly. 
  - CMake complains abour missing development targets from many of the python modules on LC. 
- We need `decorator` (for `PYB11Generator`), as well as `sphinx` and `sphinx-rtd-theme`. 
	- These are stored in `scripts/build-requirements.txt`
- Using the new `Spheral_Python_Env` function we build a virtual environment in the build tree with these dependencies. All `pybind11` code and documentation is generated from this environment.
	- **TODO**: PYB11Generators CMake call needs to take the name of a virtual environment target for the c++ generation to depend on. It will need to extract the *activation* command and change out the *python_exe* variable provided from the venv target properties.
#### Runtime Deps
- Using the same `Spheral_Python_Env` function we can generate a Virtual Env in the Install directory from `runtime-requirements.txt`.
- `spheral-setup-venv` will now copy only the necessary Spheral libraries into this environment at the install time.

The build and runtime environment target scripts will be run on every `make` call. However after the initial installation on the first `make` this only takes a few seconds as pip will check each time to ensure the environments satisfy their respective requirements.txt files.

### Caching Builds for LC
It seems as if all pip packages install from a local Livermore pypi repo. This could be very advantageous and might mean we may not have to manage cached tars ourselves.
- ATS will be a problem on air-gapped systems, currently it needs to pull ATS from the github repo.

### C++ Only Builds
- When building Spheral w/o the python interface we no longer need spack to build python.
	- Note: It will still try as caliper has a hard dependency on python.
	- **TODO**: Remove Caliper python dependency in caliper package? (this will get rid of python from our spack build tree entirely for `~python` specs).
------
### ToDo :

### DIFF
--- a/cmake/spheral/SpheralAddLibs.cmake
+++ b/cmake/spheral/SpheralAddLibs.cmake
@@ -175,7 +175,6 @@ function(spheral_add_pybind11_library package_name module_list_name)
   # List directories in which spheral .py files can be found.
   set(PYTHON_ENV 
       ${EXTRA_PYB11_SPHERAL_ENV_VARS}
-      "${BUILDTIME_PYTHONENV_STR}:"
       "${SPHERAL_ROOT_DIR}/src/PYB11:"
       "${SPHERAL_ROOT_DIR}/src/PYB11/${PYB11_MODULE_NAME}:"
       "${SPHERAL_ROOT_DIR}/src/PYB11/polytope:"
@@ -226,11 +225,20 @@ function(spheral_add_pybind11_library package_name module_list_name)
   get_property(SPHERAL_BLT_DEPENDS GLOBAL PROPERTY SPHERAL_BLT_DEPENDS)
   list(APPEND SPHERAL_DEPENDS Spheral_CXX ${${package_name}_DEPENDS})
 
+  set(PYTHON_EXE_BAK ${PYTHON_EXE})
+
+  
+  get_target_property(PYTHON_EXE python_build_env EXECUTABLE)
+
+  message("---------")
+  message("${PYTHON_EXE}")
+
+
   set(MODULE_NAME Spheral${package_name})
   PYB11Generator_add_module(${package_name}
                             MODULE          ${MODULE_NAME}
                             SOURCE          ${package_name}_PYB11.py
-                            DEPENDS         ${SPHERAL_CXX_DEPENDS} ${EXTRA_BLT_DEPENDS} ${SPHERAL_DEPENDS}
+                            DEPENDS         ${SPHERAL_CXX_DEPENDS} ${EXTRA_BLT_DEPENDS} ${SPHERAL_DEPENDS} #python_build_env
                             PYTHONPATH      ${PYTHON_ENV_STR}
                             INCLUDES        ${CMAKE_CURRENT_SOURCE_DIR} ${${package_name}_INCLUDES} ${PYBIND11_ROOT_DIR}/include
                             COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CXX>:${SPHERAL_PYB11_TARGET_FLAGS}>"
@@ -239,6 +247,8 @@ function(spheral_add_pybind11_library package_name module_list_name)
                             INSTALL         OFF
                             )
   target_include_directories(${MODULE_NAME} SYSTEM PRIVATE ${SPHERAL_EXTERN_INCLUDES})
+
+  set(PYTHON_EXE ${PYTHON_EXE_BAK})
 
   install(TARGETS     ${MODULE_NAME}
           DESTINATION ${SPHERAL_SITE_PACKAGES_PATH}/Spheral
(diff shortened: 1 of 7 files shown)


cmake/spheral/SpheralPRT.cmake
```diff
@@ -0,0 +1,125 @@
...
+    add_custom_command(
+      OUTPUT ${${target_name}_PREFIX}/.venv/${target_name}_stamp
+      COMMAND ${Python3_EXECUTABLE} -m venv --system-site-packages ${${target_name}_PREFIX}/.venv;
+      COMMAND . ${${target_name}_PREFIX}/.venv/bin/activate &&
+
+      ${PIP_INSTALL_CMD} ${REQUIREMENTS_ARGS}
+
+      DEPENDS Python3::Python ${${target_name}_REQUIREMENTS}
+    )
+  else()
```
mdavis36, line 53: blueos will not be supported as of this PR - Remove this

cmake/spheral/SpheralAddLibs.cmake
```diff
@@ -237,6 +236,7 @@ function(spheral_add_pybind11_library package_name module_list_name)
                             USE_BLT         ON
                             EXTRA_SOURCE    ${${package_name}_SOURCES}
                             INSTALL         OFF
+                            VIRTUAL_ENV     python_build_env
```
mdavis36, line 239: Need to make a Pyb11Generator PR to support this change.

extern/PYB11Generator
mdavis36, line 1: Merge changes to Develop then update submodule here.

scripts/spack/configs/toss_4_x86_64_ib_cray/compilers.yaml
```diff
@@ -0,0 +1,27 @@
...
+        cc: /usr/tce/packages/cce-tce/cce-18.0.0/bin/craycc
+        cxx: /usr/tce/packages/cce-tce/cce-18.0.0/bin/crayCC
+        f77: /usr/tce/packages/cce-tce/cce-18.0.0/bin/crayftn
+        fc: /usr/tce/packages/cce-tce/cce-18.0.0/bin/crayftn
+      flags: {}
+      operating_system: rhel8
+      target: x86_64
+      modules: []
+      environment: {}
+      extra_rpaths: []
```
mdavis36, line 27: Remove, this isn't used

Dockerfile
```diff
@@ -58,23 +74,26 @@ ARG HOST_CONFIG=docker-$SPEC
...
+WORKDIR /home/spheral/workspace/
+
 # Copy Spheral source and generate host config from tpl-manager (all dependencies should already be installed).
 COPY . .
 RUN python3 scripts/devtools/tpl-manager.py --spec $SPEC --upstream-dir /home/spack/opt/spack/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder__/__spack_path_placeholder_ --spack-url /home/spack
 
 # Configure Spheral with SPEC TPLs.
 RUN mv *.cmake $HOST_CONFIG.cmake
-RUN python3 scripts/devtools/host-config-build.py --host-config $HOST_CONFIG.cmake
+RUN python3 scripts/devtools/host-config-build.py --host-config $HOST_CONFIG.cmake -DSPHERAL_NETWORK_CONNECTED=Off
```
mdavis36, line 85: Add documentation on `SPHERAL_NETWORK_CONNECTED`

RELEASE_NOTES.md
```diff
@@ -57,6 +57,7 @@ Notable changes include:
     * Bugfix for RZ solid CRKSPH with compatible energy.
     * Parsing of None string now always becomes None python type. Tests have been updated accordingly.
     * IO for checkpoints and visuzalization can now be properly turned off through SpheralController input options.
+    * Bugfix for atomicWeight in ANEOS
```
mdavis36, line 61: Add RELEASE_NOTES Documentation

cmake/spheral/SpheralPRT.cmake
```diff
@@ -0,0 +1,125 @@
+#----------------------------------------------------------------------------------------
+#                                   Spheral_Python_Runtime_Env
+#----------------------------------------------------------------------------------------
+
+
+set(SPHERAL_PIP_CACHE_DIR ~/.cache/spheral_pip)
```
mdavis36, line 6: Document `SPHERAL_PIP_CACHE_DIR`

mdavis36 (commented):
I can't approve this since I'm the author, but I think this is ready for a final review @ldowen @jmikeowen

jmikeowen (approved):

PR #317, opened 2024-12-04
Splitting Artificial viscosity out of hydrodynamics packages into its own physics package

jmikeowen:
# Summary

- This PR is a feature/refactoring
- It does the following:
  - Make the ArtificialViscosity it's own physics package, working alongside the hydrodynamics package.
  - Cleans up some state usage, particularly being more rigorous about some of the artificial viscosity logic correctly time centering it's dependent state during a time advancement cycle.

------
### ToDo :


ptsuji:
@jmikeowen - I'd like to test this branch against the SpheralC stuff. We will need the HydroFieldNames::viscousWork variable reinstated.

src/SPH/CMakeLists.txt
```diff
@@ -37,16 +37,13 @@ endif()
...
-    SPHHydroBase.hh
-    SPHHydroBaseInline.hh
-    SPHHydroBaseRZ.hh
-    SphericalSPHHydroBase.hh
-    SolidSPHHydroBase.hh
-    SolidSPHHydroBaseInline.hh
-    SolidSPHHydroBaseRZ.hh
-    SolidSphericalSPHHydroBase.hh
+    PSPH.hh
+    SPHBase.hh
```
ptsuji, line 41: The SpheralC stuff needs SPH.hh here.

src/Utilities/DeprecationWarning.hh
```diff
@@ -0,0 +1,23 @@
+//---------------------------------Spheral++----------------------------------//
```
ptsuji, line 1: This header needs to be added to the CMakeLists.txt file, otherwise the SpheralC stuff is missing the header.

src/SPH/CMakeLists.txt
jmikeowen: Good catch -- fixed

src/Utilities/DeprecationWarning.hh
jmikeowen: Good catch -- fixed

src/Neighbor/PairwiseFieldElementAccessor.hh
```diff
@@ -0,0 +1,43 @@
+//---------------------------------Spheral++----------------------------------//
```
ptsuji, line 1: I had to include the PairwiseField.hh header to SpheralPseudoScript to resolve a "error: implicit instantiation of undefined template" error; I also had to add this header to the CMakeLists.txt file so that it would be installed.

src/Utilities/StrideIterator.hh
```diff
@@ -0,0 +1,35 @@
+//---------------------------------Spheral++----------------------------------//
```
ptsuji, line 1: I had to add this header as well to the CMakeLists.txt file so that it would be installed, after including PairwiseField.hh.

src/Neighbor/PairwiseFieldElementAccessor.hh
jmikeowen: Fixed

src/Utilities/StrideIterator.hh
jmikeowen: Fixed this one too

src/ArtificialViscosity/ArtificialViscosity.cc
jmpearl, line 1: ArtificialViscosity is a child of the handle class but calls Physics() instead of ArtificialViscosityHandle() in its constructor and then constructs some of the fieldLists but not all the fieldLists it inherits from the handle class. Looks like Handle owns EffViscPressure but ArtificialViscosity doesn’t build that one? I just haven’t seen this pattern before so double checking this right?

src/CRKSPH/CRKSPHBase.hh
jmpearl, line 1: Should effViscPressure and maxViscousPressure get stripped out of here?

src/GSPH/GSPHEvaluateDerivatives.cc
```diff
@@ -64,19 +65,19 @@ evaluateDerivatives(const typename Dimension::Scalar time,
...
-  auto  newRiemannDvDx = derivatives.fields(ReplaceState<Dimension, Scalar>::prefix() + GSPHFieldNames::RiemannVelocityGradient,Tensor::zero);
+  const auto  M = derivs.fields(HydroFieldNames::M_SPHCorrection, Tensor::zero);
+  const auto  DrhoDx = derivs.fields(GSPHFieldNames::densityGradient, Vector::zero);
+  auto  normalization = derivs.fields(HydroFieldNames::normalization, 0.0);
+  auto  DxDt = derivs.fields(IncrementState<Dimension, Vector>::prefix() + HydroFieldNames::position, Vector::zero);
+  auto  DrhoDt = derivs.fields(IncrementState<Dimension, Scalar>::prefix() + HydroFieldNames::massDensity, 0.0);
+  auto  DvDt = derivs.fields(HydroFieldNames::hydroAcceleration, Vector::zero);
+  auto  DepsDt = derivs.fields(IncrementState<Dimension, Scalar>::prefix() + HydroFieldNames::specificThermalEnergy, 0.0);
+  auto  DvDx = derivs.fields(HydroFieldNames::velocityGradient, Tensor::zero);
+  auto* pairAccelerationsPtr = derivs.template getPtr<PairAccelerationsType>(HydroFieldNames::pairAccelerations);
```
jmpearl, line 76: The pattern for accessing the pairFieldLists is a little different here then in FSISPH?

jmpearl: same for MFM and MFV

src/GSPH/MFVEvaluateDerivatives.cc
```diff
@@ -105,12 +105,8 @@ secondDerivativesLoop(const typename Dimension::Scalar time,
...
   CHECK(newRiemannDpDx.size() == numNodeLists);
   CHECK(newRiemannDvDx.size() == numNodeLists);
-
-  if (compatibleEnergy){
-    pairAccelerations.resize(npairs);
-    pairDepsDt.resize(2*npairs);
-    pairMassFlux.resize(npairs);
-  }
+  CHECK((compatibleEnergy and pairAccelerationsPtr->size() == npairs) or not compatibleEnergy);
+  CHECK(not compatibleEnergy or pairDepsDt.size() == npairs);
```
jmpearl, line 109: Should we add a check for pairMassFlux similar ot the pairDepsDt one?

src/ArtificialViscosity/ArtificialViscosity.cc
jmikeowen: This is in the ArtificialViscosity.cc, which should have been deleted.  All the implementation for the ArtificialViscosity class is in the header now, since it's pretty slim.  I've now deleted the errant ArtificialViscosity.cc.

src/CRKSPH/CRKSPHBase.hh
jmikeowen: Yep!  Good catch.

src/GSPH/GSPHEvaluateDerivatives.cc
jmikeowen: The pattern using pointers is necessary for cases where these fields are not present in the state.  The non-pointer usage in some of these classes is a mistake, which I've now fixed. This means however we must not be testing these hydros in non-compatible form.  Maybe we should be?

src/GSPH/MFVEvaluateDerivatives.cc
jmikeowen: Yep -- another good catch (fixed).

jmpearl (approved):

PR #331, opened 2025-02-07
Introducing Spack environments

ldowen:
# Summary

- This PR is a refactor of how Spack is used by Spheral
- It does the following:
  - Removes all use of uberenv.
  - Adds Spack environments.
  - Centralizes TPL designations like upstream location.
  - Adds logic to make installing Spheral TPLs outside of LC easier:
     - Uses `spack external find` and `spack compiler find` to avoid installing existing TPLs on the system.

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#129)


scripts/lc/install-from-dev-pkg.sh
```diff
@@ -28,7 +28,7 @@ mkdir -p $INSTALL_DIR
 
 cp -a $PWD/resources/pip_cache/. $SPHERAL_PIP_CACHE_DIR
 
-./$SCRIPT_DIR/devtools/tpl-manager.py --spack-url $SPACK_URL --init-only --spec=none --no-upstream --spheral-spack-dir $INSTALL_DIR/spheral-spack-tpls
+./$SCRIPT_DIR/devtools/tpl-manager.py --spack-url $SPACK_URL --init-only --no-upstream --spack-dir $INSTALL_DIR/spheral-spack-tpls
```
mdavis36, line 31: This call to TPL manager was initially for cloning and setting up a spack repository on an air-gapped system, it didn't try to concretize or install anything. Since the new default for tpl-manager without spec is to install all dependencies what happens here now?

mdavis36 (changes requested):

scripts/lc/install-from-dev-pkg.sh
ldowen: The `--init-only` flag means it only clones a spack repo so it should do what we want here.

docs/developer/dev/submodules.rst
```diff
@@ -1,7 +1,7 @@
 Submodules
 ##########
 
-Spheral uses submodules to rbing other codes/project directly into our source tree, such as BLT, and Uberenv. This guide explains some of the common practices one might want to perform with submodules during Spheral development.
+Spheral uses submodules to rbing other codes/project directly into our source tree, such as BLT, and PYB11Generator. This guide explains some of the common practices one might want to perform with submodules during Spheral development.
```
mdavis36, line 4:
```suggestion
Spheral uses submodules to bring other codes/project directly into our source tree, such as BLT, and PYB11Generator. This guide explains some of the common practices one might want to perform with submodules during Spheral development.
```

.gitlab/scripts.yml
```diff
@@ -169,7 +169,7 @@
     - tar -xzf $DEV_PKG_NAME.tar.gz
     - cd $DEV_PKG_NAME
 
-    - env INSTALL_DIR=$INSTALL_DIR SPEC=$SPEC SPACK_PKG_NAME=$SPACK_PKG_NAME BUILD_ALLOC="" SCRIPT_DIR=$SCRIPT_DIR
+    - env INSTALL_DIR=$INSTALL_DIR DEV_PKG_SPEC=$DEV_PKG_SPEC SPACK_PKG_NAME=$SPACK_PKG_NAME BUILD_ALLOC="" SCRIPT_DIR=$SCRIPT_DIR
```
mdavis36, line 172: The DEV_PKG_NAME env variable previously was the name of the final tar file we stored, it looks like we're still extracting DEV_PKG_NAME.tar.gz in this gitlab script, however in `install-from-dev-pkg.sh` we are defining the name of the tar file as DEV_PKG_SPEC, is this gitlab script wrong or is `install-from-dev-pk`, or am I missing something else here?

scripts/lc/install-from-dev-pkg.sh
```diff
@@ -17,31 +18,36 @@ if [[ -z "${INSTALL_DIR}" ]]; then
...
 echo $SCRIPT_DIR
 echo $BUILD_ALLOC
 
-rm -rf $INSTALL_DIR
-mkdir -p $INSTALL_DIR
+DEV_TAR_FILE=$DEV_PKG_SPEC.tar.gz
+
+# Delete the install directory but not the tar file, this way we can rerun this stage if a failure happens
+# like with the network
+find $INSTALL_DIR -type f ! -name '$DEV_TAR_FILE' -delete
```
mdavis36, line 31: Here.

mdavis36 (changes requested):
A couple of questions first but looks really good

.gitlab/scripts.yml
ldowen: Yeah I think this is an error. Good catch.

.gitlab-ci.yml
```diff
@@ -1,15 +1,22 @@
+workflow:
+  auto_cancel:
+    on_new_commit: interruptible
+  rules:
+    - if: $CI_MERGE_REQUEST_ID
+      when: never
+    - when: always
+
```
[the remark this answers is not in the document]
ldowen: The auto_cancel is the thing to stop running pipelines if a new commit comes through. The bottom rule is to prevent the CI from launching the same pipeline multiple times. It might only be necessary if a single pipeline is launched from multiple repos so we probably don't need it.

mdavis36 (approved):
LGTM

jmikeowen (approved):
Looks good to me

PR #333, opened 2025-02-18
Feature/cull pairs

brbass:
### Summary

This PR does two things: 

- Adds a simpler sum density for FSISPH
- Adds an option to remove connectivity for given points

------
### ToDo :

### DIFF
--- a/src/FSISPH/SolidFSISPHEvaluateDerivatives.cc
+++ b/src/FSISPH/SolidFSISPHEvaluateDerivatives.cc
@@ -233,6 +233,10 @@ secondDerivativesLoop(const typename Dimension::Scalar time,
       j = pairs[kk].j_node;
       nodeListi = pairs[kk].i_list;
       nodeListj = pairs[kk].j_list;
+      if (pairs[kk].f_couple < 1.0e-16)
+      {
+         continue;
+      }
 
       // Get the state for node i.
       const auto& interfaceFlagsi = interfaceFlags(nodeListi,i);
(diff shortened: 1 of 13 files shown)


src/FSISPH/SolidFSISPHEvaluateDerivatives.cc
```diff
@@ -233,6 +233,10 @@ secondDerivativesLoop(const typename Dimension::Scalar time,
       j = pairs[kk].j_node;
       nodeListi = pairs[kk].i_list;
       nodeListj = pairs[kk].j_list;
+      if (pairs[kk].f_couple < 1.0e-16)
+      {
```
jmpearl, line 237: I believe this will break fsi. Depending upon the node coupling algorithm, f_couple can be zero at damage interfaces and this will give non-physical behavior when compression happens along a damage interface. FSI has it's own decoupling logic to handle damage fronts.

What was the need for the additional decoupling?

brbass: I removed it! Was trying something new when debugging.

src/Neighbor/ConnectivityMap.cc
```diff
@@ -1139,6 +1147,8 @@ computeConnectivity() {
   //   }
   // }
 
+  mNodePairListPtr->makeUnique();
```
brbass, line 1150: I don't remember why I added this, but it might not be needed.

src/CRKSPH/CRKSPH.cc
```diff
@@ -89,6 +89,15 @@ CRKSPH(DataBase<Dimension>& dataBase,
...
 
+//------------------------------------------------------------------------------
+// Destructor.
+//------------------------------------------------------------------------------
+template<typename Dimension>
+CRKSPH<Dimension>::
+~CRKSPH(){
+  // Needs to be here due to implicit PairwiseField delete
+}
+
```
brbass, line 100: Had to add this to all the hydro packages. Otherwise, the build fails since the PairwiseField destructor isn't available in the header file.

PR #335, opened 2025-03-24
Ale3d

ptsuji:
# Summary

- Merge back ALE3D changes

------
### ToDo :


src/Integrator/Integrator.cc
jmikeowen, line 1: We're using CXXONLY here as a stand-in for something specific to ALE3D that seems like a side-effect for CXXONLY.  Could we instead introduce a new configuation flag (something like "ENABLE_GLOBALDT_REDUCTION") that defaults true, which an ALE3D build can then trip to false to specify this behavior?

adayton1: And prefix the option name with SPHERAL_ (e.g. SPHERAL_ENABLE_GLOBALDT_REDUCTION).

jmikeowen: Sure, that works for me

src/Physics/GenericHydro.cc
jmikeowen, line 1: The fix for the globaldt configuration switch above is fine, and I think we should do the same thing here with the longitudinal sound speed.  Rather than turning this criterion off as a side effect of CXXONLY, let's make a CMake flag that indicates we're changing this.  Something like SPHERAL_ENABLE_LONGCS_TIMESTEP

Gotta love long flag names...

adayton1 (approved):

jmikeowen (approved):

PR #337, opened 2025-04-09
Improve external builds/update polytope

ldowen:
# Summary

- This PR updates polytope and the documentation regarding building Spheral
- It does the following:
  - Uses the find_package logic on polytope
  - Improves the documentation for building Spheral
  - Improvements to tpl-manager and host-config-build scripts

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#131)


jmikeowen (approved):
Looks good to me -- nice job updating the docs!

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 # CMakeLists to build the Spheral library.
 
-cmake_minimum_required(VERSION 3.14)
```
mdavis36, line 3: Is there an explicit cmake 3.21 feature we need to bump this?

docs/build_guide/external/index.rst
```diff
@@ -1,19 +1,131 @@
...
+   .. tab-set::
+
+      .. tab-item:: Version 8
+
+         .. code-block::
+
+            # Required packages
+            dnf update
+            dnf install epel-release
+            dnf group install "Development Tools"
```
mdavis36, line 65:
```suggestion
            dnf groupinstall "Development Tools"
```

docs/build_guide/external/index.rst
```diff
@@ -1,19 +1,131 @@
...
+            module load mpi
+
+      .. tab-item:: Version 9
+
+         .. code-block::
+
+            # Required packages
+            dnf update
+            dnf install epel-release
+            dnf group install "Development Tools"
```
mdavis36, line 80:
```suggestion
            dnf groupinstall "Development Tools"
```

mdavis36 (changes requested):
I tried this out today on a fresh Fedora 42 install. I was unable to install the Spheral dependencies through spack:

1. Fedora 42 is on Python 3.13. Our current version of Spack fails to bootstrap clingo. (Since this isn't necessarily our fault I manually built and installed Python 3.11 to get past this).
2. `dnf` on Fedora seems to install packages in 3 places e.g. w/ cmake `/bin/cmake`, `/usr/bin/cmake` and `/usr/sbin/cmake`. Spack successfully finds all 3, however the paths it stores in the `yaml` file are (`/bin`, `/usr`, `/usr/sbin`).  During the builds spack tries to build with the path `/usr/sbin/bin/cmake` which does not exist and causes all packages that build with cmake to fail.
3. spack isn't picking up some packages I would expect `ncurses`, `curl`, `wget`, `zlib` many of which are later failing to build with the `gcc@15.1.1` for Fedora 42.

This is a shame because it seems most of these issues are out of our control and are problems with spack / it's package recipes.

There are a few improvements that I think we can address in the near future:
1. When building externally it's not always clear which packages spack has found and exactly where it is pointing to them. It would be nice to have some output communicating this.
2. I think instead of letting spack implicitly find everything on the system we should define a set of packages we always expect to be on the system like `ncurses` and explicitly search for them (like we do for python perkl and mpi right now), especially since we have users install some of those packages with their respective package managers.

This PR is a step in the right direction for external users but we aren't there yet unfortunately. I don't think we need to address any of the above right now. We should update to the spack 1.0 and see where things settle, but we will probably need to work with the spack team to resolve some of the above.

CMakeLists.txt
adayton1: You'll need CMake 3.21 to work with Hip and the latest BLT release.

mdavis36: Probably but this PR isn't bringing in any HIP compiled code, we tried bringing this version up from 3.14 to 3.18 a while ago and had issues with BLT and CUDA. The consensus after speaking with the BLT folks was to revert back to 3.14 until we had a necessity to push the version. When HIP executing changes get pushed to Spheral is probably when we will revisit our minimum version constraint.

docs/build_guide/external/index.rst
mdavis36: resolved offline

mdavis36 (approved):

PR #342, opened 2025-04-29
Gtest Suite Infrastructure.

mdavis36:
# Summary

- This PR is a feature:
  - It brings in the gtest suite changes used in #334 as a standalone change set.

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#136)


scripts/spack/packages/spheral/package.py
```diff
@@ -67,8 +67,11 @@ class Spheral(CachedCMakePackage, CudaPackage, ROCmPackage):
...
     depends_on('axom@0.9.0 +hdf5 -lua -examples -python -fortran', type='build')
-    depends_on('axom +shared', when='~cuda', type='build')
-    depends_on('axom ~shared', when='+cuda', type='build')
+    with when('+rocm') or when('+cuda'):
+        depends_on('axom ~shared', type='build')
+
+    with when('~rocm') or when('~cuda'):
+        depends_on('axom +shared', type='build')
 
     with when("+caliper"):
```
mdavis36, line 76: @ldowen do you have a preferred way to do this, my solution here seems subpar...

tests/cpp/Field/field_tests.cc
jmikeowen, line 1: It may seem a bit pedantic, but I think it's helpful to put a comment at the start of each of these tests stating what they are testing.

mdavis36: Agree, I added some basic comment blocks. Most of the tests in this PR are pretty self explanatory but once more complex behaviors start to appear with the GPU port it'll be helpful to define the expected behavior.

scripts/spack/packages/spheral/package.py
ldowen: I couldn't find a better way to do it. I think I would prefer listing them all out with `when` statements instead of using the `with when` but how you have it is fine if it works.

jmikeowen (dismissed):
Looks good to me

PR #370, opened 2025-06-25
[GPU] FieldView for heterogeneous GPU systems.

mdavis36:
# Summary

- This PR contains features to allow migration of field data on heterogeneous GPU systems.
- It does the following:
  - Sets the foundation for #371 by implementing `FieldView` datatypes as handles to be used for migrating data to and from the GPU of `Field` containers.
  - Sets up a series of unit tests for ensuring our Spheral data types such as `GeomVector<Dim>` can be copied to and from the device correctly.
  - Brings in the necessary changes to ensure `GeomVector<Dim>` can be used on the GPU #374  and tests it's functionality.

- Compiling parts of Spheral that pass through the device compiler seem to throw warnings we don't see in our current Spheral builds these have been squashed and a separate PR #387 has been made extracting these changes that should be merged immediately before this one.

------
### ToDo :

- [ ] Annotate ``RELEASE_NOTES.md`` with notable changes.
- [ ] Create LLNLSpheral PR pointing at this branch. (PR#)
- [ ] LLNLSpheral PR has passed all tests.

Issue #372, opened 2025-06-25
[GPU] FieldView


mdavis36:
We need thin wrappers that can take a handle to host data within a field and manage the duties of allocating, deallocating and copying that data to and from the device as needed. The type must:
- Use reference semantics to be compatible with RAJA lambda execution.
- NOT take ownership of host side data. (no host alloc / dealloc)
- Provide facilities to allow a user to free GPU data **explicitly**.
- Match enough of the `Spheral::Field` interface to cover the basic needs of Field access and data manipulation ( no resizing / realloc capabilities).

PR #386, opened 2025-07-10
[GPU] Geom3Vector

mdavis36:
# Summary

- This PR is a refactoring of Geom3Vector:
  - Annotates Geom3Vector with SPHERAL_HOST_DEVICE attributes macro.
  - Adds CPU & GPU unit tests for Geom3Vector.
  - Make Geom3Vector trivially copyable for use with `FieldView`.
 
- It directly addresses the requirements of #375
- It depends on #370  

------
### ToDo :


ldowen:
> > Maybe we should add some tests where we construct vectors on the host and ensure they are the same on the device. Then modify them on the device and make sure they are the same on the host.
> 
> We do that in [this test](https://github.com/LLNL/spheral/blob/d94db2c91157db10df327a31f7fc1399beeb6afd/tests/cpp/Field/fieldview_datatype_tests.cc#L43) defined in #372

Oh ok. I suppose that covers it then.

Issue #381, opened 2025-06-30
Restart safety of porosity models


jmpearl:
Porosity models are not restart safe when the user circumvents node generation on restarts. This is b/c c0 is initialized in the constructor and is note dumped/restored as part of the restart methods.

PR #394, opened 2025-07-20
v2025.06.1 Patch Release

mdavis36:
# Summary

- This PR is a for a patch release:
- It does the following:
  - Adding c0 (reference sound speed) as one of the restart variables for the porosity models.  Previously we depended on this being set during problem initialization every time, but sometimes users want to skip problem reinitialization when it's very expensive.
  - Fixes issue #381 

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. ([PR#161](https://rzlc.llnl.gov/gitlab/owen/LLNLSpheral/-/merge_requests/161))


jmikeowen (approved):

ldowen (approved):

src/Porosity/PalphaPorosity.cc
```diff
@@ -158,50 +158,59 @@ evaluateDerivatives(const Scalar time,
...
       fDSnew(i) = 1.0;
 
     } else {
 
-      // First compute the derivative with respect to pressure
-      auto DalphaDpi = 0.0;
-      if (alphai > 1.0) {
-        if (Pi < mPe or dPsdti < 0.0) {
+      // Time evolution of the solid pressure
+      const auto dPsdti = dPsdri*DrhoDti + dPsdui*DuDti;
```
jmpearl, line 170: It looks like DrhoDti is the rate of change of the bulk density and dPsDri is the derivative of pressure with respect to the solid density. If my interpretation is correct, is there an issue with the density mismatch in the partial derivatives? Should there be an alpha term in here or something like that?

jmpearl (approved):
Mike take a look at the dPsdt comment, if its a non issue then everything else looked good to me.

src/Porosity/PalphaPorosity.cc
jmikeowen: You might be right here (I have to think about it a bit more), but I'm inclined to leave this one as is for this patch (it's the same as the code prior to this patch).  Only the sign of this term matters (it's one criterion for checking if we should be in the elastic regime), and it's subtle.  I'd like to look at whether this matters in it's own change rather than this patch release.

PR #406, opened 2025-08-25
Add linear solvers

brbass:
# Summary

This PR adds linear solvers to Spheral. This supersedes the previous PR (https://github.com/LLNL/spheral/pull/346), which was inefficient in both the matrix fill and factorization. 
- Any data needed for the matrix fill needed to be requested for each row instead of once. 
- The factorization was being done every time the matrix was solved rather than being done once when the matrix is done being filled. 
- When filling the vectors with data, a copy needed to be made instead of just giving the pointer to the first element of the Field. 
- When the connectivity did not change but the data did, the entire matrix was scrapped, causing a costly reallocation of memory.

The basic usage of these new solvers is:
 - Call `initialize` to create the matrices and vectors
 - Start filling the matrix with `beginFill`
 - Fill the matrix all at once with `setMatRows` or by row with `setMatRow`
 - Call `assemble` when fill is complete to assemble the matrix
 - Call `finalize` after assembly to create solvers and preconditioners
 - Each time the solver will be used:
   - Set the RHS to the source and optionally the LHS to the guess with `set`
   - Call `solve` or `multiply`
   - Retrieve the LHS with `get`

This PR probably depends on and should be merged after https://github.com/LLNL/spheral/pull/405 due to the changes to `FlatConnectivity` . 
 
------
### ToDo :

- [X] Annotate ``RELEASE_NOTES.md`` with notable changes.

### DIFF
--- a/cmake/InstallTPLs.cmake
+++ b/cmake/InstallTPLs.cmake
@@ -194,7 +194,7 @@ set_property(GLOBAL PROPERTY SPHERAL_FP_DIRS ${SPHERAL_FP_DIRS})
 
 message("-----------------------------------------------------------------------------")
 # Use find_package to get Sundials
-if (ENABLE_SOLVERS)
+if (SPHERAL_ENABLE_SOLVERS)
   set(SUNDIALS_DIR "${sundials_DIR}")
   find_package(SUNDIALS REQUIRED NO_DEFAULT_PATH
     COMPONENTS kinsol nvecparallel nvecmpiplusx nvecserial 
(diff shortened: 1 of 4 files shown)


ldowen:
Could you change `ENABLE_SOLVERS` to be `SPHERAL_ENABLE_SOLVERS`?

brbass:
> Also, would it be possible to add tests for just `FlatConnectivity` independent of the solvers?

Yes, good call. `FlatConnectivity` was originally added along with the `KernelIntegrator` and is also tested there, but it would be good to make sure it's working correctly on its own!

ldowen:
I had a few questions for https://github.com/LLNL/spheral/pull/405 but nothing big.

brbass:
I'm going to close this and create a new version with some improvements.

Issue #412, opened 2025-09-26
Test CMake export logic in CI


ldowen:
Add tests to our CI that build and import a `ENABLE_STATIC_CXXONLY` Spheral build to ensure the export logic is not broken.

Issue #363, opened 2025-06-18
Change ENABLE_CXXONLY to SPHERAL_ENABLE_PYTHON


mdavis36:
The use of `ENABLE_CXXONLY` can be a counterintuitive, as we often check for `if (NOT ENABLE_CXXONLY)` / `#ifndef ENABLE_CXXONLY` throughout the codebase.

We should only be using this guard when we are not building the python bindings. Guarding Python binding portions of the code with `if (SPHERAL_ENABLE_PYTHON)` (or something to that effect) reads a lot clearer to users and developers.

PR #413, opened 2025-09-26
CMake variable overhaul

ldowen:
# Summary

- This PR is a refactoring of our CMake variable and export system.
- It does the following:
  - Removes `CXXONLY` flags in favor of a `SPHERAL_ENABLE_PYTHON`.
  - Prepended many CMake variables with `SPHERAL_`.
  - Removed unused variables and definitions.
  - Moved many command line variable definitions to config.hh.
  - Builds with python disabled default to building with static C++ libraries.
  - Add tests of our CMake export system to the CI (unfinished).
  - Addresses https://github.com/LLNL/spheral/issues/363 and https://github.com/LLNL/spheral/issues/412

------
### ToDo :


jmikeowen (approved):
Loks good to me

cmake/CMakeDefinitions.cmake
```diff
@@ -92,10 +77,17 @@ endif()
 include(CheckIncludeFileCXX)
 CHECK_INCLUDE_FILE_CXX(span HAS_STD_SPAN_HEADER)
 
-if(HAS_STD_SPAN_HEADER AND CMAKE_CXX_STANDARD GREATER_EQUAL 20)
+if(SPHERAL_UNIFIED_MEMORY)
```
jmikeowen: The choice of which span (std::span vs boost::span) is a separate question from whether or not we're using SPHERAL_UNIFIED_MEMORY, so I don't this this check should be restricted to that case.

PR #414, opened 2025-10-01
Collection of changes/fixes from Brody

jmikeowen:
# Summary

This PR fixes some issues and adds some new functions to existing classes.

- Fixes a few compilation warnings
- Adds a few types to the state
  - std::vector of boundary, for when you want to register the boundaries in the state
  - Option to enroll copies of items, without which enrolling a double or any other built-in type from the python interface will fail (with the standard enroll, it will create a temporary double, enroll the address of that temporary double, and then kill the temporary double)
- Adds timing to a few high-cost functions
- Updates to FlatConnectivity:
  - Adds some missed variable initializations
  - Adds unique index computation for global connectivity, without which indices would need to be summed directly
  - Fixes non-functional python bindings
- Fixes particularly awful bug where variables from an exec statement won't be loaded into globals() if a temporary variable is used elsewhere
  - If you use [x for x in range(5)] anywhere after the exec statement, it will find that x was also used as a variable in exec(from Spheral1d import *) and then import all the variables into a temporary namespace that is thrown away
  - Also fixes the above for the output function in python

------
### ToDo :


ldowen (dismissed):
LGTM as long as the TIME_FUNCTION calls aren't going overboard.

jmikeowen:
Landon already looked it over so we're good to go.

Issue #303, opened 2024-08-28
Johnson-Cook Damage Model dependency


jmpearl:
The Johnson-Cook damage model currently causes a circular dependency issue. It looks like some change (who knows how recently?) broke that damage model. Down the road, we should add a ci tests to construct and run a cycle w/ our various eos/strength/damage models just to make sure they don't get outdated like this.

Sean Stokes encountered this bug and here's the error message he sent me:

`...at line 259 of file /usr/WS2/sphapp/gitlab-runner/builds/zcqvTfGz/000/gitlab/spheral/spheral/toss_4_x86_64_ib-spheral-dev-pkg-9848/src/./DataBase/State.cc.
 
    control.advance(goalTime, maxSteps)
  File "/usr/gapps/Spheral/toss_4_x86_64_ib/spheral-9848/.venv/lib/python3.9/site-packages/Spheral/SpheralController.py", line 368, in advance
    self.integrator.step(goalTime)
RuntimeError: Verification failed: State::update ERROR: someone has specified a circular state dependency.
Remaining State:
   flaws
   plastic strain
   pressure
   shear modulus
   sound speed
   tensor damage
   yield strength
State dependencies:
H|Air : 
H|Impactor : 
H|Plate : 
bulk modulus|Air : mass density  ++  porosity solid mass density  ++  specific thermal energy  ++  
bulk modulus|Impactor : mass density  ++  porosity solid mass density  ++  specific thermal energy  ++  
bulk modulus|Plate : mass density  ++  porosity solid mass density  ++  specific thermal energy  ++  
deviatoric stress|Air : 
deviatoric stress|Impactor : 
deviatoric stress|Plate : 
flaws|Impactor : deviatoric stress  ++  melt specific energy  ++  plastic strain  ++  pressure  ++  specific thermal energy  ++  
flaws|Plate : deviatoric stress  ++  melt specific energy  ++  plastic strain  ++  pressure  ++  specific thermal energy  ++  
interfaceAreaVectors|Air : 
interfaceAreaVectors|Impactor : 
interfaceAreaVectors|Plate : 
interfaceFlags|Air : 
interfaceFlags|Impactor : 
interfaceFlags|Plate : 
interfaceNormals|Air : 
interfaceNormals|Impactor : 
interfaceNormals|Plate : 
interfaceSmoothness|Air : 
interfaceSmoothness|Impactor : 
interfaceSmoothness|Plate : 
mass density|Air : 
mass density|Impactor : 
[...]

PR #429, opened 2025-11-14
EOS models, unit conversion, particle fixups

brbass:
# Summary

### Singularity EOS
  - Includes GammaLawGas, ExtendedVinet, MieGruneisen, Spiner, and EOSPAC.
  - Tested against the ordinary GammaLawGas in spheral.
  - Will improve a lot once spack is updated, as the current singularity version is a few years out of date.

### Null EOS
- `NullEquationOfState` is the counterpart of `NullStrength`.

### Unit conversion
- Converts between two instances of `PhysicalConstants`. Very helpful for things like Singularity above. 

### Subdirectories in silo
- Support for subdirectories in silo files. Currently used for RK corrections. 

### Improvements for the smoothing parameter
  - Adds a node separation and relaxation method that slots into the initial `iterateIdealH` calculations. To use it, create `relaxation = RelaxNodes(...)` and then pass it to the controller: `SpheralController(initialHExtraPackages = [relaxation])`. This separates nodes that are too close and relaxes nodes using the Voronoi, without which the smoothing parameter may not converge for problems with problematic initial node distributions. 
  - Adds a check for the initial H distribution to limit the number of initial points to some reasonable value. If a problem passes in nodes that have hundreds of thousands of neighbors or whatever, this sharply limits the number before the `iterateIdealH` calculations get underway. 
  - Adds a damping factor for the IdealH calculation that helps it converge when struggling.

### Verbose time step control
- Adds a `verboseStep` parameter to the `Integrator`. Printing the verbose time step info every cycle makes things difficult to read. With the verbose step set to 100, it is reasonable to keep it on all the time. 

## Bugfixes
-  Removed the old, non-functional volume calculation in `SPH` and requested Voronoi cells instead. 
- Removed circular dependencies for Johnson-Cook to let it run. https://github.com/LLNL/spheral/issues/303. This will lag the strain calculation. 
- Added `allReduceLoc` to prevent the time step from being printed once per rank when they all calculate the same time step.
- Added a `gradientPairs`  method that is more robust for points with difficult H distributions (yes, more smoothing parameter issues). Also cleaned up the `gradient` instantiation.
- Added some missing includes.

## ToDo :


ldowen:
Help me understand what is going on here. It seems like singularity has a ton of dependencies but I don't understand why we have the Spheral Spack package depending on Singularity's dependencies and why we are bringing them into our CMake. Neither of those things should be happening. It is possible to use a newer Spack recipe for Singularity than what is in the current version of Spack we are using, if that will help you. You might have to tweak it a little.

ldowen:
Based on the singular-eos Spack package recipe, Kokkos is an optional variant. Unless this is a mistake, it implies that Singularity can work without Kokkos. By not including `+kokkos +kokkos-kernel` on your `depends_on('singularity-eos` the version of singularity you are building will have Kokkos turned off. Again, unless there is a mistake in the package recipe in Spack.

brbass:
Have you guys started doing internal tests for this PR? 

Working in 1D with the gnuplot files is onerous, so I made the siloPointmeshDump compatible with 1D and made that the default (which can be reverted if anyone is actually using the gnuplot version in production). It depends on the changes to siloMeshDump in this PR. Would you prefer that I add them here or create a separate PR that depends on this one?

jmikeowen:
There is an existing Spheral1dViz method that outputs the data as a table of numbers.  This should be pretty friendly to all sorts of 1D plotting packages -- what's different in your version?

brbass:
> There is an existing Spheral1dViz method that outputs the data as a table of numbers. This should be pretty friendly to all sorts of 1D plotting packages -- what's different in your version?

This outputs it in silo format, so you can load the curve data directly in visit. Then you can create movies or whatever using the same workflow as in 2D/3D. Here is an example:

image

jmikeowen:
Ah I see.  Personally I don't like using Visit for 1D plots but I can understand this would be useful for folks that do.  I'm happy to have this as an option for 1D viz output, but I don't think it should be the default.

jmikeowen:
Another question on a different change: the RelaxNodes package added for iterateIdealH. It appears to me this duplicates the relaxNodeDistribution functionality we have in NodeGenerators, which is invoked by SpheralController optionally.  We might want to rearrange how that's used a bit to fit your use case here?  I'd rather not have iterateIdealH also adjust the point positions, which seems like an unexpected side-effect of converging the smoothing scale distribution.

brbass:
Sure, we don't need to make visit the default in 1D. 

The reason RelaxNodes is optionally included in the H iterations is just convenience, since iterateIdealH already includes the Voronoi evaluation and other initializations. From what I can see, prerelaxNodeDistribution doesn't call relaxNodeDistribution but instead calls an hourglass control package. The relaxNodeDistribution function doesn't seem to be tested or used anywhere except testStar2d.py, which isn't in the testing suite. It uses an old mesh object rather than the Voronoi cells. All that said, there's no reason to add RelaxNodes to the repo if you don't think it will be generally useful. It meets some of our needs for separating and differentiating points when generating them from a non-ideal distribution, but that isn't a situation anyone is likely to encounter when using the built-in generators. 

It seems like the easiest path forward is:
- Include the visit visualization files for 1D but leave gnuplot as the default
- Remove RelaxNodes, since it may not be generally useful outside of niche applications

ldowen:
@brbass We merged the Spack 1.1 update yesterday. Do you want to revisit this PR?

brbass:
I just finished merging that in! I would propose that we close this MR and I can replace it with several smaller ones for bugfixes, singularity, and so on, unless that would make life harder for you guys. 

I could also just update this MR with more recent changes to singularity. Let me know what you would prefer

jmikeowen:
Smaller merges are almost certainly easier!

brbass:
Sounds good. Many MRs incoming later today!

Issue #442, opened 2025-12-04
Update TPL Manager to use spack v1.1.0


mcfadden8:
The current version of `tpl-manager.py` uses v0.22.0 of `spack`.  This PR is to document work done in support of this update.

### External Dependencies
- [x] #441 - (needed by tool chains built by spack v1)
### Work
- [x] Update spack packages to use new `cmake` and/or `autoconf` `spack builder classes` available in spack v0.22.0.  This is needed because many/most of the older mechanisms used are difficult to maintain or port in `spack v1.0+`
- [x] Update `yaml` files to match with documented style changes required by `spack`
- [x] Update packages to be compliant with new spack package API (2.0) that is used by `spack 1.1`.

mcfadden8:
Fixed: This was fixed as part of the upgrade to spack v1 PR #421 (Docker image: 24.04, gcc 13.3.0, python3 3.12.3)

PR #443, opened 2025-12-04
Add ATS Updates

mcfadden8:
# Summary
Update ATS to work with later versions of python (needed for #442)


mcfadden8:
This has already been addressed with #444

PR #462, opened 2026-01-31
Creating View class for PairwiseField

jmikeowen:
# Summary

- This PR is a feature
- It does the following:
  - Breaks up the PairwiseField using View/Value pattern into PairwiseFieldView and PairwiseField, such that PairwiseFieldView is appropriate for use on GPU devices.

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#196)

### DIFF
--- a/src/Neighbor/PairwiseFieldView.hh
+++ b/src/Neighbor/PairwiseFieldView.hh
@@ -1,0 +1,81 @@
+//---------------------------------Spheral++----------------------------------//
+// PairwiseFieldView -- view class for PairwiseField
+//
+// Stores a value per node pair in a NodePairList. Because connectivity is
+// allowed to change step to step in our meshfree methods, PairwiseField is
+// ephemeral and will be invalidated when topology is updated.
+//
+// Created by J. Michael Owen, Wed Nov 20 14:44:44 PST 2024
+//----------------------------------------------------------------------------//
+#ifndef _Spheral_NeighborSpace_PairwiseFieldView_
+#define _Spheral_NeighborSpace_PairwiseFieldView_
+
+#include "Neighbor/PairwiseFieldElementAccessor.hh"
+#include "Utilities/DataTypeTraits.hh"
+#include "Utilities/StrideIterator.hh"
+#include "Utilities/DBC.hh"
+
+#include "chai/ManagedArray.hpp"
+#include "chai/ExecutionSpaces.hpp"
+
+#ifdef SPHERAL_UNIFIED_MEMORY
+#include "Utilities/span.hh"
+#endif
+
+namespace Spheral {
+
+// Forward declarations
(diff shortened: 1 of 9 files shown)


tests/cpp/Neighbor/pairwisefieldview_tests.cc
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    // Check the values in a RAJA loop
+    RAJA::forall<WORK_EXEC_POLICY>(TRS_UINT(0, N),
+       [=] SPHERAL_HOST_DEVICE (size_t i) {
+         SPHERAL_ASSERT_EQ(dfield_v[i], 0.0);
+         SPHERAL_ASSERT_EQ(ddfield_v[i][0], 0.0);
+         SPHERAL_ASSERT_EQ(ddfield_v[i][1], 1.0);
+       });
+
```
ldowen, line 110: Should we check the counters here with COMP_COUNTERS like we do with other tests like FieldView?

src/Neighbor/PairwiseFieldView.hh
```diff
@@ -0,0 +1,81 @@
...
+
+  // Comparators
+  SPHERAL_HOST_DEVICE bool operator==(const PairwiseFieldView& rhs) const           { return mSpan == rhs.mSpan; }
+  SPHERAL_HOST_DEVICE bool operator!=(const PairwiseFieldView& rhs) const           { return mSpan != rhs.mSpan; }
+
+  // Other methods
+  SPHERAL_HOST_DEVICE size_t size() const                                           { return mSpan.size()/numElements; }
+
+  // Zero the Field
+  SPHERAL_HOST_DEVICE void Zero()                                                   { for (auto& x: mSpan) x = DataTypeTraits<Value>::zero(); }
```
ldowen, line 71: I'm not sure if this matters but not all of the DataTypeTraits zero functions are set as host device.

jmikeowen: Hmm.  Good point.  Since this is an inline method that's not hitting us since the Value types we're instantiating do have those methods host/device.  I'd say let's leave this as is, and if we introduce new Fields (or PairwiseFields) of types that need that changed we look at it then.

tests/cpp/Neighbor/pairwisefieldview_tests.cc
jmikeowen: I didn't add those callback methods to the PairwiseFieldView since I didn't want to introduce that sort of complication (and extra storage) to the class.  Do we want those callback furnctions for every view class?

ldowen: Oh right! No that's fine.

ldowen (approved):

PR #471, opened 2026-02-19
Fix broken schedules and redundant lib installs

### LINKED ISSUE #427: Unnecessary duplicate python files/libraries installed
Our CMake is installing the Spheral python files and libraries twice. Once in .venv/lib/python3.9/site-packages/Spheral and again in lib/python3.9/site-packages/Spheral. The second install is unused and should not occur.

ldowen:
# Summary

- This PR is a bug fix.
- It does the following:
  - Removes redundant library installs.
  - Fixes the broken performance pipeline schedule.
  - Fixes broken viz with matplotlib by switching to python3+tkinter and using the system installs for tk and tcl.
  - Moves adiak routines out of PYB11Generator preamble into a separate file.

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#200)


.gitlab/scripts.yml
```diff
@@ -129,6 +130,7 @@
   stage: update_tpls
   rules:
     - !reference [.merge_pr_rule, rules]
+    - if: '$CI_COMMIT_MESSAGE =~ /update-tpls/'
```
mcfadden8, line 135: Should/Can this trigger and its intended usage be documented someplace?  What is the expected result when someone adds this comment in one of their own activity branches?

ldowen: We have a few of these things. I will add some comments to the scripts file to mention what they do. This one should simply run the update tpls pipeline meaning any new tpl versions are installed in the upstream.

mcfadden8 (dismissed):

jmikeowen (dismissed):

PR #475, opened 2026-03-04
Miscellaneous bugfixes and cleanup

brbass:
# Summary

- Removes the old, non-functional volume calculation from SPH and requests Voronoi cells from RK instead. 
- Adds missing includes across several files. 
- Moves NodeIteratorBase stream operators from the inline header to the source file. 
- Fixes the Johnson-Cook damage model to avoid circular dependencies
- Adds parallel GenerateRatioSphere in 2D. 
- Includes material internal energy in the generic node distribution. 

------
### ToDo :


mcfadden8:
[build_fail.txt](https://github.com/user-attachments/files/27249538/build_fail.txt)

Hi @brbass - Our CI is reporting a build failure (attached snippet).  Have you seen this before?

brbass:
> [build_fail.txt](https://github.com/user-attachments/files/27249538/build_fail.txt)
> 
> Hi @brbass - Our CI is reporting a build failure (attached snippet). Have you seen this before?

Fixed! I think it is probably a difference of compilers since this warning didn't cause an error in my local build.

jmikeowen:
We have some CI builds enforcing warnings as errors just to keep things clean, so that's probably the difference.

src/KernelIntegrator/FlatConnectivity.cc
```diff
@@ -883,6 +884,8 @@ getUniqueIndices(const std::vector<unsigned>& globalNeighbors,
   std::map<unsigned, unsigned> globalToIndex;
   for (auto j = 0u; j < size; ++j) {
     const auto global = globalNeighbors[j];
+    CHECK(global < numGlobalNodes());
```
ldowen: There is also a mismatched type comparison here that only shows up when built with debug. @brbass

brbass: Fixed two of those!

src/Damage/JohnsonCookDamagePolicy.cc
```diff
@@ -26,8 +26,7 @@ namespace Spheral {
 template<typename Dimension>
 JohnsonCookDamagePolicy<Dimension>::
 JohnsonCookDamagePolicy():
-  UpdatePolicyBase<Dimension>({SolidFieldNames::flaws,
-                               SolidFieldNames::plasticStrain}) {
+  UpdatePolicyBase<Dimension>() {
```
jmikeowen, line 29: Why remove these dependencies?  I believe this policy does depend on both the flaws and plastic strain -- was there a circular dependency resulting?

Also as an aside the JC damage and strength are not well tested, so be cautious using them.

src/FSISPH/SolidFSISPH.cc
```diff
@@ -31,7 +31,7 @@
 #include "DataBase/State.hh"
 #include "DataBase/StateDerivatives.hh"
 #include "DataBase/IncrementState.hh"
-#include "DataBase/ReplaceBoundedState.hh"
+#include "DataBase/ReplaceState.hh"
```
jmikeowen, line 34: I see this was a redundant include with one a few lines later, but in fact I think FSI doesn't use ReplaceState at all.  Maybe this include can just be dropped.

src/SPH/SolidSPH.cc
```diff
@@ -185,7 +185,7 @@ initializeProblemStartupDependencies(DataBase<Dimension>& dataBase,
   TIME_BEGIN("SolidSPHinitializeStartup");
 
   // Call the ancestor.
-  SPHBase<Dimension>::initializeProblemStartupDependencies(dataBase, state, derivs);
+  SPH<Dimension>::initializeProblemStartupDependencies(dataBase, state, derivs);
```
jmikeowen, line 188: This one is a C++ question for me: the SPH class does not have an initializeProblemDependencies method, but it's ancestor (SPHBase) does.  What happens by calling the SPH:: declaration here?

mcfadden8:
Hi @brbass, we are hopeful to get this into our next release.   I'm ready to take any changes you might have in response to the questions/comments raised in this PR.  If no changes are required, please let me know that as well.

Thanks!

src/Damage/JohnsonCookDamagePolicy.cc
brbass: https://github.com/llnl/spheral/issues/303

src/SPH/SolidSPH.cc
brbass: If there is no SPH::declaration, then it will call SPHBase::declaration. But if there is an SPH::declaration in the future, this will call that instead. Seems like an awful bug to track down, so this is just being nice to our future selves.

brbass:
> Hi @brbass, we are hopeful to get this into our next release. I'm ready to take any changes you might have in response to the questions/comments raised in this PR. If no changes are required, please let me know that as well.
> 
> Thanks!

Responded to all! Feel free to merge whenever you guys think it is ready.

mcfadden8:
#495 merged - Closing (and thanks!)

PR #476, opened 2026-03-04
Improvements to integrator output

brbass:
# Summary

- Adds an allReduceLoc utility so the controlling time step is printed once with its owning rank, rather than once per rank, when the time steps do not vary by proc. 
- Adds verbose time step frequency. This allows the code to print out every N time steps, meaning the verbose time step can always be on but not flood the output. 
- Allows periodic work frequencies in SpheralController to be callables, so they can change dynamically during a simulation.

------
### ToDo :


mcfadden8:
Merged with #488

PR #480, opened 2026-03-04
Smoothing scale improvements

brbass:
# Summary

- Adds a damping factor to the ideal-H iteration in both SPH and ASPH smoothing scales, which helps convergence for difficult initial distributions. 
- Adds a check in SpheralController that limits the initial neighbor count to a reasonable value before iterateIdealH begins. 
- Supports extra packages during initial H iteration and a flag to force Voronoi tessellation.

------
### ToDo :


src/SimulationControl/SpheralController.py
```diff
@@ -902,13 +909,55 @@ def dropViz(self,
         print("Wrote viz file in %0.2f seconds" % (time.time() - start))
         return
 
+    #--------------------------------------------------------------------------
```
jmikeowen, line 912: What is the purpose of this method?  It appears to be forcing H to be smaller if there are more points than would be implied if the kernelExtent was 1 and points are perfectly arranged.  Not all kernels have kernelExtent=1 though.

brbass: If you start off with a large density gradient, for instance, a single point can encompass tens of thousands of others. This starts of iterateIdealH with a reasonable number of neighbors. If you don't think this function will be generally useful, we can remove it. If we want to keep it, then we should include the kernelExtent.

mcfadden8: @jmikeowen - let me know if there is anything specific you would like to see changed with respect to this (I'm not seeing anything from this note at the moment).  thanks..

mcfadden8:
Merged with #488

PR #484, opened 2026-03-09
Fixed typo in ShadowIvanoviSALEDamageModel.py file and added guard to JohnsonCookStrength

swiggins8:
Found a typo in the ShadowIvanoviSALEDamageModel.py file (was kward now kwarg). Also added an additional guard to the JohnsonCookStrength model so that nobody puts a 0 for the minimum stress rate.

# Summary

- This PR is bugfix and a guard addition to a strength model
- It does the following:
  - Fixes a typo in ShadowIvanoviSALEDamageModel.py
     - Typo was "kward_order" when it should have been "kwarg_order"
     - This could be a problem if the user wanted to have multiple positional arguments.  
  - Added an additional guard to the JohnsonCookStrength model
     - Before the user could enter a value less than or equal to zero for the minimum allowed strain rate. 
     - This would cause -inf values
     - New guard prevents the user from running the strength model if their input isn't greater than zero.

------
### ToDo :

### DIFF
--- a/src/SolidMaterial/JohnsonCookStrength.cc
+++ b/src/SolidMaterial/JohnsonCookStrength.cc
@@ -47,6 +47,8 @@ JohnsonCookStrength(const SolidEquationOfState<Dimension>& eos,
   mShearModulusScaling(shearModulusScaling) {
   VERIFY2(mEpsdot0 > 0.0,
           "JohnsonCookStrength ERROR: reference strain-rate must be greater than zero.");
+  VERIFY2(mEpsdotmin > 0.0,
+          "JohnsonCookStrength ERROR: reference strain-rate must be greater than zero.");
   VERIFY2(mTmelt > mTroom,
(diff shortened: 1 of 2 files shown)


src/SolidMaterial/JohnsonCookStrength.cc
```diff
@@ -47,6 +47,8 @@ JohnsonCookStrength(const SolidEquationOfState<Dimension>& eos,
   mShearModulusScaling(shearModulusScaling) {
   VERIFY2(mEpsdot0 > 0.0,
           "JohnsonCookStrength ERROR: reference strain-rate must be greater than zero.");
+  VERIFY2(mEpsdotmin > 0.0,
+          "JohnsonCookStrength ERROR: reference strain-rate must be greater than zero.");
```
jmpearl: update the error message and I think this should be good to go.

jmpearl (changes requested):

jmpearl (approved):

jmikeowen (approved):

PR #485, opened 2026-03-11
Update TPLs and versions for GPUs

### LINKED ISSUE #483: [GPU] Update TPL packages and tests
Update the following:
- Update to ROCM 6.4.3.
- Add more external system libraries for LC configs.
- Update RAJA perf suite to v2025.12.0.
- Spack package repo to newer commit to allow for newer RAJA version.
- Ensure the Spack package repo commit is updated when the tpl manager runs.
- Update ATS tests to ensure only rank per GPU is used.

ldowen:
# Summary

- This PR is an update.
- It does the following:
  - Updates versions of ROCM, CHAI, Umpire, RAJA, Boost, and LEOS to prepare upcoming GPU changes.
  - Updates the spack package hash accordingly.
  - Includes some general CI fixes.

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#201)

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -40,7 +40,7 @@ Notable changes include:
       * Added actual Spheral release versions and TPL conditions to Spheral package.
       * Added more system libraries to make building on LC faster.
     * TPL updates:
-      * LEOS updated from 4.2.0 to 5.2.0.
+      * LEOS updated from 8.4.2 to 8.5.2.
       * ROCM version updated from 6.2.0 to 6.4.3.
       * Boost updated from 1.85 to 1.87 to prepare for using Clang 20 eventually.
       * CHAI, Umpire, and RAJA updated to 2025.12.0.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -31,6 +31,7 @@ Notable changes include:
       * Gathering of performance data for the deploy stage is now parallelized over ranks and threads.
       * Merges to develop will create installs to a shared directory for use by the performance testing scripts.
       * Added umask command for updating upstreams and removed separate job to update permissions.
+      * Build and test job now fails if import of Spheral module fails.
     * Builds and installs are cleaner:
       * Rpaths are no longer overwritten, allowing things set in the Spack host config file to be used.
       * Spheral libraries are only installed once now and a Spheral.pth with a relative path to the install lib is used in the virtual python environment.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,6 +20,9 @@ Notable changes include:
       so there is no user interface impact.
     * Added update-tpls commit message trigger.
     * Spheral mpi python interface ensures proper allocation calls are used on Flux machines to avoid strange hangs that can occur if running outside of allocations.
+    * tpl-manager.py updates:
+      * Added --no-upstream option for when on LC machines but cannot access upstream.
+      * Added --dry-run option for testing purposes.
     * Performance testing and CI improvements:
       * Structure for performance tests is different, with the tests themselves set in a perf_tests.py file and run_perf.py runs the perf tests using ATS.
       * The cleanup old directories job is now a scheduled pipelines that runs weekly.
@@ -27,6 +30,7 @@ Notable changes include:
       * Gitlab pages now uses Plotly for interactive visualizations.
       * Gathering of performance data for the deploy stage is now parallelized over ranks and threads.
       * Merges to develop will create installs to a shared directory for use by the performance testing scripts.
+      * Added umask command for updating upstreams and removed separate job to update permissions.
     * Builds and installs are cleaner:
       * Rpaths are no longer overwritten, allowing things set in the Spack host config file to be used.
       * Spheral libraries are only installed once now and a Spheral.pth with a relative path to the install lib is used in the virtual python environment.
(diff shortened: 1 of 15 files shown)


scripts/devtools/tpl-manager.py
```diff
@@ -309,9 +315,7 @@ def concretize_spec(self, check_spec):
         else:
             print("Concretizing environment")
 
-        with self.spack_env.write_transaction():
-            conc_spec = self.spack_env.concretize(force=force_conc)
-            self.spack_env.write()
+        conc_spec = self.do_concretize(force_conc)
```
mcfadden8, line 318: I was under the impression that this needed to be synchronized which was why I added the `write_transaction`.  I believe that it resolved an issue for me that only occurred occasionally.  But I cannot remember exactly what that was.  (I think it may have been an error when/if the subsequent `spack_env.concrete_roots()` was called)

RELEASE_NOTES.md
```diff
@@ -20,20 +20,30 @@ Notable changes include:
...
     * Builds and installs are cleaner:
       * Rpaths are no longer overwritten, allowing things set in the Spack host config file to be used.
       * Spheral libraries are only installed once now and a Spheral.pth with a relative path to the install lib is used in the virtual python environment.
     * Updated build system support spack v1.
       * Updated TPL Manager to support spack v1.1.0.
       * Updated CI to support python v3.12 and Ubuntu 24.04.
       * Added actual Spheral release versions and TPL conditions to Spheral package.
+      * Added more system libraries to make building on LC faster.
+    * TPL updates:
+      * LEOS updated from 4.2.0 to 5.2.0.
```
ptsuji: I think you mean updating LEOS from 8.4.2 to 8.5.2 here?

ldowen: Yep good catch. Thanks.

scripts/devtools/tpl-manager.py
ldowen: It worked for me locally. Think you could find out when it would fail?

scripts/spack/environments/dev_pkg/spack.yaml
```diff
@@ -10,8 +10,14 @@ spack:
     spheral:
       path: ../../../../
       spec: spheral@=develop
-  include::
-    - ../../configs/toss_4_x86_64_ib/packages.yaml
-    - ../../configs/config.yaml
+  'include:':
```
mcfadden8, line 13: I'm never sure about the world of yaml and spack's use of the extra ":", but I am curious why this particular line changed?

scripts/spack/environments/dev_pkg/spack.yaml
```diff
@@ -10,8 +10,14 @@ spack:
...
+  - ../../configs/toss_4_x86_64_ib/packages.yaml
+  - ../../configs/config.yaml
   repos:
-    - ../../spack_repo/spheral
+    spheral: ../../spack_repo/spheral
+  packages:
+    all:
+      permissions:
```
mcfadden8, line 20: Is this section really necessary?  Isn't this taken care of by the `umask` and sticky-bit of the containing directory where this is being placed?

scripts/spack/environments/dev_pkg/spack.yaml
ldowen: The changes to the environment files were done directly by spack when I had spack locally modify the environment. I decided to leave the changes in since they appear to be how spack likes the yaml files to look.

scripts/spack/environments/dev_pkg/spack.yaml
ldowen: It should be but I figured it can't hurt since that package is only used for updating the upstream.

scripts/devtools/tpl-manager.py
mcfadden8: Yes, it would work for me locally too.  But then sometimes it would fail.  I think I was able to reproduce it most often on rzwhippet when I would run without a `--spec` argument to cause it concretize more.

ldowen: Oh I think I realized what you mean. I didn't remove the write_transaction call. I just wrapped it up in a `do_concretize` call that is defined a few lines down. Is that what you mean?

mcfadden8: Oh duh.  `self.do_concretize` is **NOT** the same as `self.spack_env.concretize`.  Nevermind.

mcfadden8 (dismissed):

scripts/devtools/tpl-manager.py
```diff
@@ -330,6 +334,15 @@ def concretize_spec(self, check_spec):
             specs = self.spack_env.concrete_roots()
             self.print_specs(specs)
 
+    def do_concretize(self, force_conc):
+        if (self.args.no_upstream):
+            spack.config.set("upstreams", {"spheral_shared": {"install_tree": {}}})
+            print("WARNING: Modifying local Spack files, do not commit these changes.")
```
mcfadden8, line 340: Not necessarily for this PR, but rather than warning the user, I think it would be better to change the spack files, if really needed, then restore them before returning control to the user.

scripts/devtools/tpl-manager.py
```diff
@@ -57,10 +57,14 @@ def parse_args(self):
                             help="Skip downloading Spack repo.")
         parser.add_argument("--ci-run", action="store_true",
                             help="For use by the CI only. Must set a --spec.")
+        parser.add_argument("--no-upstream", action="store_true",
```
jmikeowen, line 60: Minor point, but I'd make the help line here more along the lines of "Force local build of TPL's, ignoring any upstream prebuilt packages"
There might be reasons for doing this even if the upstream is not broken, who are we to judge?  :-)

jmikeowen (dismissed):

jmikeowen (approved):

PR #478, opened 2026-03-04
Silo improvements

brbass:
# Summary

- Supports subdirectories in silo files.
- Permits output to silo for 1D, which can be read in the Curves in VisIt. 

------
### ToDo :

- [ ] Annotate ``RELEASE_NOTES.md`` with notable changes.
- [ ] Create LLNLSpheral PR pointing at this branch. (PR#)
- [ ] LLNLSpheral PR has passed all tests.

PR #477, opened 2026-03-04
Added pairwise gradient and improved instantiations

brbass:
# Summary

- Adds gradientPairs to FieldListFunctions, which computes the gradient using node pairs.
- Cleans up the gradient instantiation template.

------
### ToDo :

- [ ] Annotate ``RELEASE_NOTES.md`` with notable changes.
- [ ] Create LLNLSpheral PR pointing at this branch. (PR#)
- [ ] LLNLSpheral PR has passed all tests.

PR #474, opened 2026-03-04
Insert axis boundary in controller, not hydro

brbass:
# Summary

Moves the axis boundary logic out of individual SPH and CRKSPH hydro constructors and into the SpheralController, where it is applied once during initialization. This makes the axis BC work without hydro, if needed.

------
### ToDo :

- [ ] Annotate ``RELEASE_NOTES.md`` with notable changes.
- [ ] Create LLNLSpheral PR pointing at this branch. (PR#)
- [ ] LLNLSpheral PR has passed all tests.

PR #488, opened 2026-03-25
Pr 474 476 477 478 480 combo

mcfadden8:
# Summary
Combined PRS: #474 #476 #477 #478 and #480 
- This PR is a (refactoring, bugfix, feature, something else)
- It does the following:
  - Modifies / refactors (class or method) (how?)
  - Fixes (issue number(s))
  - Adds (specific feature) at the request of (project or person)

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#203)


mcfadden8:
Hi @brbass - Thanks for these PRs.  I'm helping to bring these PRs into spheral and we have opted to combine these 5 PRs into this one.  The internal tests are passing.  Would you be able to send in an update to the RELEASE_NOTES.md file with anything notable concerning these?

Thanks,

Marty

brbass:
> Hi @brbass - Thanks for these PRs. I'm helping to bring these PRs into spheral and we have opted to combine these 5 PRs into this one. The internal tests are passing. Would you be able to send in an update to the RELEASE_NOTES.md file with anything notable concerning these?
> 
> Thanks,
> 
> Marty

Hi Marty! I don't think I can edit this MR, so I have attached my proposed additions to the release notes. 
[RELEASE_NODES.md](https://github.com/user-attachments/files/26281666/RELEASE_NODES.md)

src/SimulationControl/SpheralController.py
```diff
@@ -902,13 +936,55 @@ def dropViz(self,
...
+        pos = dataBase.globalPosition
+        conn = dataBase.connectivityMap()
+        nodeLists = dataBase.nodeLists
+        numNodeLists = dataBase.numNodeLists
+
+        numLimited = 0
+        for ni in range(numNodeLists):
+            # Get the approximate goal number of nodes for this NodeList
+            nPerh = nodeLists[ni].nodesPerSmoothingScale
+            pi = 3.14159
```
jmikeowen: This should defintely be removed and just use Python's pi value instead

src/FieldOperations/gradient.cc
```diff
@@ -277,6 +278,60 @@ gradient(const FieldList<Dimension, std::vector<DataType>>& fieldList,
   return result;
 }
 
+//------------------------------------------------------------------------------
+// Calculate the gradient of a FieldList.
+//------------------------------------------------------------------------------
+template<typename Dimension, typename DataType>
+void
+gradientPairs(FieldList<Dimension, typename MathTraits<Dimension, DataType>::GradientType>& result,
```
ldowen, line 286: @brbass Would it make sense to try to thread this function like we do in eval derivs? Can we add tests for it to ensure it doesn't break at some point?

src/PYB11/Silo/SiloWrappers.hh
```diff
@@ -1392,7 +1392,7 @@ DBPutPointmesh(DBfile& file,
 
   // Preconditions.
   const unsigned ndims = coords.size();
-  VERIFY(ndims == 2 or ndims == 3);
+  VERIFY(ndims == 1 or ndims == 2 or ndims == 3);
```
ldowen, line 1395: @brbass Can we add tests for 1D silo stuff if that functionality is being added?

src/FieldOperations/gradient.cc
brbass: When I spoke to Mike last about the threading situation, he said not to bother with it since the way threading is done might change with GPU work. If that interface has settled down, it certainly wouldn't hurt.

src/Distributed/allReduce.hh
```diff
@@ -27,22 +27,45 @@ namespace Spheral {
...
           const MPI_Comm comm = Communicator::communicator()) {
+  CHECK(!(op == SPHERAL_OP_MINLOC || op == SPHERAL_OP_MAXLOC));
   Value tmp = value;
   Value result;
   MPI_Allreduce(&tmp, &result, 1,
                 DataTypeTraits<Value>::MpiDataType(), op, comm);
   return result;
 }
 
+template<typename Value>
```
jmikeowen, line 45: Warning for the future -- this method is returning the index as an integer, but we've been trying to more index types to size_t in Spheral (int's have bitten us a couple of times already for this).  This is complicated here because the native MPI MINLOC uses ints, which is frustrating.  We can define our own versions of these operations and force size_t, which we may want to consider soon if not right away in this PR.  @brbass

src/SimulationControl/SpheralController.py
brbass: Yes, agreed! I don't think I can edit this branch, so feel free to make the change on your end.

src/Distributed/allReduce.hh
brbass: As this value is compared to `Process::getRank()`, which returns an integer, I would say we leave it alone. If you change `Process::getRank()` return `size_t`, then this function will trigger a warning and can be changed then.

src/SimulationControl/SpheralController.py
```diff
@@ -902,13 +936,55 @@ def dropViz(self,
...
+            if self.dim == 1:
+                goalN = 2.0 * nPerh
+            elif self.dim == 2:
+                goalN = pi * nPerh ** 2
+            else:
+                goalN = 4.0 / 3.0 * pi * nPerh ** 3
```
brbass:
```suggestion
            if self.dim == 1:
                goalN = 2.0 * nPerh
            elif self.dim == 2:
                goalN = np.pi * nPerh ** 2
            else:
                goalN = 4.0 / 3.0 * np.pi * nPerh ** 3
```

mcfadden8:
@brbass , @jmikeowen @ldowen - I think it would be most prudent for us to continue this commentary in Brody's individual PRs so that he can make the changes locally and I can pull them in as comments get marked as satisfactorily resolved.

src/PYB11/Silo/SiloWrappers.hh
brbass: I don't think there are dedicated visualization tests. Since the Noh tests have visualization except for in 1D, I would suggest that would be a good place to at least exercise the code. Something like this is all that would be needed:

```python
commandLine(
    ...
    vizCycle = None,
    vizTime = 0.1,
    ...
)
...
vizDir = os.path.join(dataDir, "visit")
...
if vizTime is None and vizCycle is None:
    vizBaseName = None
else:
    vizBaseName = restartBaseName
...
from SpheralPointmeshSiloDump import dumpPhysicsState
control = SpheralController(..., 
                            vizMethod = dumpPhysicsState,
                            vizBaseName = vizbaseName,
                            vizDir = vizDir,
                            vizStep = vizCycle,
                            vizTime = vizTime,
                            ...)
```

ldowen: Oh right. This is the viz stuff. Yeah sounds like that is something we should look into adding at some point

src/SimulationControl/SpheralController.py
mcfadden8: Done.

src/Distributed/allReduce.hh
mcfadden8: @jmikeowen - I'm pushing to get this PR merged. Let me know if there something specific to be done in this PR.

jmikeowen (approved):
Good to go

ldowen (approved):

PR #502, opened 2026-05-22
Modification/time step constraint

jmpearl:
# Summary

- This PR introduces a different way to enforce the maximum-acceleration-magnitude-based time-step criterion in the GenericHydro (dx = 1/2 a dt^2). The default is still the previous approach. The user can access the new approach by setting: hydro.useNewAccelerationMagnitudeForDt=True in their script. This edit is meant to correct excessive low dt when a block of material is traveling at high velocity with low acceleration (tripping the h/vmax portion of the old acceleration magnitude criterion).

------
### ToDo :


jmpearl:
I purposefully tried to keep the edit pretty light -- i.e. i didn't propagate through all the children constructors. Happy to expose this throughout though if you think that would be better Mike.

jmikeowen:
Looks good to me.  Two things we might want to consider:
- Should your new criterion just replace the old rather than have the old option?  Does it miss anything?
- If we do keep it as an option, we should probably make your new option an argument in our hydro factory functions (SPHHydros.py, FSISPHydros.py, CRKSPHHydros.py, etc) consistent with the other timestep controls.

ldowen:
Don't forget to make an LLNLSpheral version of this.

jmikeowen (dismissed):

ldowen (approved):
LGTM

PR #503, opened 2026-05-22
ale3d

ptsuji:
# Summary

- This PR brings over changes from ALE3D's submodule branch of Spheral
  - Contains fixes for various ALE3D builds


jmikeowen (dismissed):
Looks good to me.  I'm not sure why we commented out the size_t stuff in DataTypeTraits originally, seems harmless to have if needed. We may not need to screen it out with __APPLE__, but this is fine as is for now.

src/Kernel/NSincPolynomialKernel.hh
```diff
@@ -26,6 +26,14 @@ public:
   // Constructor.
   NSincPolynomialKernel(const int order);
 
+  // ALE3D begin edit for working around a ROCm 7 bug.
```
adayton1: Remove references to application.

ptsuji: Ok, I've changed this comment.

src/Distributed/allReduce.hh
```diff
@@ -31,7 +31,7 @@ namespace Spheral {
 #define SPHERAL_OP_MAXLOC MPI_MAXLOC
 
 template<typename Value>
-constexpr Value
+inline Value
```
adayton1: What compiler complained about constexpr?

jmikeowen: Removing constexpr is correct here (I don't know when it was introduced) -- I actually have this change in a branch of mine as well.  Since this is an MPI operation the result definitely is not known at compile time, I don't know why compilers didn't complain about this before?

ptsuji: I think it was with clang 20 on my Mac laptop but can't remember for sure.

adayton1 (dismissed):

jmikeowen (approved):

adayton1 (approved):

PR #510, opened 2026-06-24
Refactor NodeList into base and view layers

mcfadden8:
# Summary
Split NodeList into NodeListBase and NodeListView modeled after the FieldBase/FieldView pattern. NodeList now inherits from both layers while retaining ownership of its fields and node mutation logic.

NodeListView exposes only the data needed for neighbor finding in CPU/GPU execution spaces: node counts, positions, and Hfield FieldViews. It follows the existing RAJA/CHAI guard style with host/device accessors and explicit move/touch handling for non-unified-memory builds.

- This PR is a refactoring endeavor
- It does the following:
  - Refactors NodeList into NodeListBase and NodeListView modeled after the FieldBase/FieldView pattern. NodeList now inherits from both layers while retaining ownership of its fields and node mutation logic.

------
### ToDo :


mcfadden8:
Using https://github.com/llnl/spheral/tree/review/mcfadden8/nodelist-value-view approach instead.

PR #516, opened 2026-07-09
Include Spack and Spack packages into buildcache logic

### LINKED ISSUE #472: Tar spack and spack package instance for buildcache
Have the dev_pkg make a different install root outside of the spheral-spack-tpls directory so that we can tar up the spheral-spack-tpls directory and use it with the build cache instead of having the install pkg script create it's own spack instance.
Also, have the --init-only command for tpl-manager load a spack environment as well to ensure it also creates a spack-package repo in the right spot.

ldowen:
# Summary

- This PR is a bugfix for the build cache logic.
- It does the following:
  - tpl-manager now creates a tar of the spack and spack packages repos alongside the dev pkg tar.
  - Generating the build cache uses this logic to create the tar file.
  - Installing the build cache uses the tar file instead of specifying the spack url.

------
### ToDo :

- [x] Create LLNLSpheral PR pointing at this branch. (PR#217)


scripts/lc/install-from-dev-pkg.sh
```diff
@@ -15,9 +14,28 @@ if [[ -z "${INSTALL_DIR}" ]]; then
   echo "INSTALL_DIR var must be set."
   exit 1
 fi
+# Check if a tar file of the spack and spack package repos exists
+SPACK_CACHE_DIR=${SPACK_CACHE_DIR:-${DEV_PKG_SPEC}-spack.tar.gz}
+if [[ -f '$SPACK_CACHE_DIR' ]]; then
```
mcfadden8:
```suggestion
if [[ -f "$SPACK_CACHE_DIR" ]]; then
```

scripts/lc/install-from-dev-pkg.sh
```diff
@@ -15,9 +14,28 @@ if [[ -z "${INSTALL_DIR}" ]]; then
...
+    TPL_DIR=$INSTALL_DIR/$(tar -tzf $SPACK_CACHE_DIR | head -1 | cut -d/ -f1)
+    if [[ ! -d "${TPL_DIR}" ]]; then
+        tar -xzf $SPACK_CACHE_DIR -C $INSTALL_DIR
+    fi
+else
+    # Otherwise, see if the TPL_DIR has been set explicitly
+    TPL_DIR=${TPL_DIR:-$INSTALL_DIR/spheral-spack-tpls}
+fi
+# Now ensure the proper directories exist
+if [[ -d "${TPL_DIR}/spack" ]]; then
```
mcfadden8: Are these inverted?
```suggestion
if [[ ! -d "${TPL_DIR}/spack" ]]; then
```

scripts/lc/install-from-dev-pkg.sh
```diff
@@ -15,9 +14,28 @@ if [[ -z "${INSTALL_DIR}" ]]; then
...
+    echo "Spack repo must exist in TPL_DIR"
+    exit 1
+elif [[ -d "${TPL_DIR}/packages" ]]; then
```
mcfadden8:
```suggestion
elif [[ ! -d "${TPL_DIR}/packages" ]]; then
```

mcfadden8 (changes requested):

mcfadden8 (dismissed):

jmikeowen (dismissed):

scripts/lc/install-from-dev-pkg.sh
```diff
@@ -16,19 +15,36 @@ if [[ -z "${INSTALL_DIR}" ]]; then
   exit 1
 fi
 
+mkdir -p $INSTALL_DIR
+
+# Check if a tar file of the spack and spack package repos exists
+SPACK_CACHE_DIR=${SPACK_CACHE_DIR:-${DEV_PKG_SPEC}-spack.tar.gz}
```
mcfadden8: Should this be DEV_PKG_NAME like in generate-buildcache.sh?
```suggestion
SPACK_CACHE_DIR=${SPACK_CACHE_DIR:-${DEV_PKG_NAME}-spack.tar.gz}
```

mcfadden8: I'm not sure whether DEV_PKG_NAME needs to be regenerated here, it only seemed to be different from what SPEC was (to me).

ldowen: Yeah good catch. It is supposed to be name but name isn't actually passed in. I think I will just change the default name altogether to simplify things.

mcfadden8 (approved):

PR #518, opened 2026-07-15
Extending damage models to work in RZ coordinates

jmikeowen:
# Summary

- This PR is a feature
- It does the following:
  - Extends the damage modeling to work in axisymmetric (RZ) coordinates
  - As part of the above, Tensor and SymTensor classes have been extended to always include the zz component (for 2D) and yy and zz components (in 1D)

------
### ToDo :


src/Damage/TensorStrainPolicy.cc
```diff
@@ -17,17 +17,119 @@
...
-#include <vector>
-using std::vector;
-using std::string;
-using std::pair;
-using std::make_pair;
+#include <functional>
 
 namespace Spheral {
 
+namespace {  // anonymous
```
ldowen: Any point to keep this huge block here?

src/Geometry/GeomTensorBase.hh
```diff
@@ -15,48 +15,69 @@ template<int nDim> class GeomTensorBase {};
...
-  SPHERAL_HOST_DEVICE GeomTensorBase(
-                 const double xx, const double xy,
-                 const double yx, const double yy):
+    mxy(0.0),
+    myx(0.0),
+    myy(a),
+    mzz(a) {}
+  SPHERAL_HOST_DEVICE GeomTensorBase(const double xx, const double xy,
+                                     const double yx, const double yy,
+                                                                       const double zz = 0.0):
```
ldowen, line 50: Missed an indent.

src/NodeList/NodeListView.hh
```diff
@@ -55,18 +55,18 @@ public:
...
-                                   const PositionView& positions,
-                                   const VelocityView& velocity,
-                                   const HfieldView& Hfield,
-                                   const WorkView& work,
-                                   const Scalar hmin = 1.0e-20,
-                                   const Scalar hmax = 1.0e20,
-                                   const Scalar hminratio = 0.1,
-                                   const Scalar nPerh = 2.01,
-                                   const size_t maxNumNeighbors = 500);
+  // SPHERAL_HOST_DEVICE NodeListView(const size_t numNodes,
```
ldowen: Maybe just remove this altogether if you don't think it is useful.

ldowen (commented):
A few cosmetic changes but overall it looks good.

src/Damage/TensorStrainPolicy.cc
jmikeowen: Naw, that can go at this point.

src/Geometry/GeomTensorBase.hh
jmikeowen: Actually that's on purpose!  It puts the zz component in the correct place if you take into account the missing (implicitly zero) xz, yz, zx, and zy components.

ldowen: Oh I see. Well I see why you did that but I can't say I like it lol.

src/NodeList/NodeListView.hh
jmikeowen: I think I commented that out when I had some problem (possibly the new compiler complained about it?)  Don't recall now, but I think in general we only need constructors that are invoked by the Value class (NodeList), so this shouldn't be needed I think.  

I meant to close the loop on this with @mcfadden8   -- Marty, is there some reason we needed this extra NodeListView constructor?

mcfadden8: Ugh.  No reason other than I missed this while cleaning up.

PR #531, opened 2026-09-21
Adding back the ability to develop ArtificlalViscosities from Python

jmikeowen:
# Summary

- This PR is a feature
- It does the following:
  - Brings back the ability to create new experimental ArtificialViscosities in Python. 
    - This feature was accidentally lost during the conversion of the ArtificialViscosities for use on the GPU.
    - Note: Python ArtificialViscosities will only work on CPU runs.

------
### ToDo :


src/ArtificialViscosity/PythonArtificialViscosity.cc
```diff
@@ -0,0 +1,132 @@
...
+
+namespace Spheral {
+
+//------------------------------------------------------------------------------
+// Constructor
+//------------------------------------------------------------------------------
+template<typename Dimension, typename QPiType>
+PythonArtificialViscosity<Dimension, QPiType>::
+PythonArtificialViscosity(const Scalar Clinear,
+                                const Scalar Cquadratic,
```
ldowen: Missed some indents here.

src/ArtificialViscosity/PythonArtificialViscosity.cc
```diff
@@ -0,0 +1,132 @@
...
+      const Vector& xj,
+      const SymTensor& Hj,
+      const Vector& etaj,
+      const Vector& vj,
+      const Scalar rhoj,
+      const Scalar csj,
+      const FieldListView<Dimension, Scalar>& fCl,
+      const FieldListView<Dimension, Scalar>& fCq,
+      const FieldListView<Dimension, Tensor>& DvDx) const {
+
```
ldowen, line 108: Do we want to ifdef an error here if called on device? Otherwise, it will just return a not very helpful seg fault. Unless there is something already protecting this that I am missing.

ldowen (commented):
Few minor things but looks good overall.

src/ArtificialViscosity/PythonArtificialViscosity.cc
jmikeowen: What should this test and fail look like to check if called on device?

ldowen: Something like:
```
#ifdef SPHERAL_GPU_ACTIVE
printf("ERROR: cannot call python artvisc class on device");
abort();
#else
```

src/ArtificialViscosity/PythonArtificialViscosity.cc
jmikeowen: Fixed

src/ArtificialViscosity/PythonArtificialViscosity.cc
jmikeowen: I've changed the build system so these Python overridable classes are not built when we're configured for GPUs, so that should take care of this problem.

ldowen (approved):
LGTM

